PR #72, opened 2022-07-20
adding a best solution for raw data

alejomonbar:
## Description
Once we find the minimum of the QAOA cost function, we can access the most probable state. However, in many cases, it is not the solution with the minimum Hamiltonian cost from the set of bitstring outcomes. The additional function, I'm adding gives the bitstring with the minimum Hamiltonian cost.
## Checklist

## Type of change

### DIFF
--- a/openqaoa/optimizers/result.py
+++ b/openqaoa/optimizers/result.py
@@ -28,47 +28,18 @@ def most_probable_bitstring(cost_hamiltonian, measurement_outcomes):
     mea_out = list(measurement_outcomes.values())
     index_likliest_states = np.argwhere(mea_out == np.max(mea_out))
     # degeneracy = len(index_likliest_states)
-    solutions_bitstrings = [list(measurement_outcomes.keys())[
-        e[0]] for e in index_likliest_states]
+    solutions_bitstrings = [
+        list(measurement_outcomes.keys())[e[0]] for e in index_likliest_states
+    ]
 
-    return {'solutions_bitstrings': solutions_bitstrings,
-            'bitstring_energy': bitstring_energy(cost_hamiltonian, solutions_bitstrings[0])}
+    return {
+        "solutions_bitstrings": solutions_bitstrings,
+        "bitstring_energy": bitstring_energy(cost_hamiltonian, solutions_bitstrings[0]),
+    }
 
-def lowest_cost_bitstrings(cost_hamiltonian:Hamiltonian, measurement_outcomes:dict, n_bitstrings:int = 1) -> dict:
+
+class Result:
     """
-    Find the minimium energy from cost_hamilonian given a set of measurement
-    outcoms
-
-    Parameters
-    ----------
-    cost_hamiltonian : Hamiltonian
-        The cost Hamiltonian of the problem.
-    measurement_outcomes : dict
-        Dictionary with keys and values equals to the bitstring the probability,
-        repectively of a quantum circuit.
-    n_bitstrings : int
-        Number of the lowest energies bistrings to get
-
-    Returns
-    -------
-    best_results : dict
-        Returns a list of bitstring with the lowest values of the cost Hamiltonian.
-
-    """
-    solution_bitstring = list(measurement_outcomes.keys())
-    energies = [bitstring_energy(cost_hamiltonian, bitstring) for bitstring in solution_bitstring]
-    args_sorted = np.argsort(energies)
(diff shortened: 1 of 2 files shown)


openqaoa/optimizers/result.py
```diff
@@ -34,6 +34,34 @@ def most_probable_bitstring(cost_hamiltonian, measurement_outcomes):
     return {'solutions_bitstrings': solutions_bitstrings,
             'bitstring_energy': bitstring_energy(cost_hamiltonian, solutions_bitstrings[0])}
 
+def lowest_cost_bitstring(cost_hamiltonian:Hamiltonian, measurement_outcomes:dict) -> dict:
```
Q-lds, line 37: I think it would be nice if this would return your lowest-k energies, where `k` is an integer which defaults to 1 but in general is specified by the user

openqaoa/optimizers/result.py
```diff
@@ -34,6 +34,34 @@ def most_probable_bitstring(cost_hamiltonian, measurement_outcomes):
...
+    -------
+    dict
+        Returns the bitstring with the lowest value of the cost Hamiltonian.
+
+    """
+    min_energy = np.inf
+    solution_bitstring = None
+    for bitstring in measurement_outcomes.keys():
+        energy = bitstring_energy(cost_hamiltonian, bitstring)
+        if energy <  min_energy:
```
Q-lds, line 60: Is this handling states with the same energy?

openqaoa/optimizers/result.py
alejomonbar: done! Thanks for your suggestion.

openqaoa/optimizers/result.py
alejomonbar: Now, the logic changes and we select n_bitstrings of the lowest energies.

openqaoa/optimizers/result.py
```diff
@@ -34,6 +34,38 @@ def most_probable_bitstring(cost_hamiltonian, measurement_outcomes):
     return {'solutions_bitstrings': solutions_bitstrings,
             'bitstring_energy': bitstring_energy(cost_hamiltonian, solutions_bitstrings[0])}
 
+def lowest_cost_bitstrings(cost_hamiltonian:Hamiltonian, measurement_outcomes:dict, n_bitstrings:int = 1) -> dict:
```
vishal-ph, line 37: perhaps it makes sense to add this function as a method of the class `Result`. Also, this way the user can call the method by passing the argument `n_bitstrings` when needed and can save computation (performed inside the __init__) when not required.

tests/test_qpu_devices.py
```diff
@@ -87,7 +87,7 @@ def test_check_connection_provider_no_backend_provided_credentials(self):
         used are correct, check_connection should return True.
         The provider_connected attribute should be updated to True.
         """
-
+        raise ValueError(f"The API_TOKEN: {self.API_TOKEN}")
```
vishal-ph, line 90: What's the purpose of this line? I think it can be located in a better position.

openqaoa/optimizers/result.py
alejomonbar: I agree. I'll do so. Do you think I should make the same for the other function `most_probable_bitstring`?

vishal-ph: I think, probably @Q-lds remembers the reason why we decided to put this function outside the `Result` class. I believe there was a reason but I can't remember.

Q-lds: I think we were using it in the tests as a in individual function (otherwise, we may turn it into a class module?)

vishal-ph: Right okay! I think let's leave that function in place for now for minimal changes through this PR.

vishal-ph:
You formatted the `result.py` using python black, it seems like it 😄, looks good! At some point, I want to format the rest of the source code as well

vishal-ph:
@Q-lds, as you mentioned already, some tests are failing because the forks don't have access to the github secret api_token. Do we already want to try a possible solution for this PR itself?

Q-lds:
No, I don't think there is a solution to that. They will be run after the merge into dev anyway

vishal-ph:
> No, I don't think there is a solution to that. They will be run after the merge into dev anyway

Alright, I'll approve the changes then

vishal-ph (approved):

Issue #48, opened 2022-06-28
Knapsack random generation error


lockwo:
When running simple code to generate a random knapsack problem, I encounter the following error. I do not know where this is coming from or if I am doing something wrong. 

Code:
```python
q = QAOA()
problem = Knapsack.random_instance(n_items=9, quantum=True).get_pubo_problem()
q.compile(problem)
q.optimize()
```

Error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [71], in <cell line: 13>()
     49 q = QAOA()
     50 problem = Knapsack.random_instance(n_items=9, quantum=True).get_pubo_problem()
---> 51 q.compile(problem)
     52 q.optimize()
     53 ks[-1].append(q.results_information["best cost"][0])

File c:\users\owenl\documents\openqaoa-main\openqaoa\workflows\optimizer.py:345, in QAOA.compile(self, problem, verbose)
    324 """
    325 Initialise the trainable parameters for QAOA according to the specified
    326 strategies and by passing the problem statement
   (...)
    339     Set True to have a summary of QAOA to displayed after compilation
    340 """
    341 self.mixer_hamil = get_mixer_hamiltonian(n_qubits=problem.n,
    342                                          qubit_connectivity=self.circuit_properties.mixer_qubit_connectivity,
    343                                          coeffs=self.circuit_properties.mixer_coeffs)
--> 345 self.cost_hamil = Hamiltonian.classical_hamiltonian(
    346     terms=problem.terms, coeffs=problem.weights, constant=problem.constant)
    348 self.circuit_params = QAOACircuitParams(
    349     self.cost_hamil, self.mixer_hamil, p=self.circuit_properties.p)
    350 self.variate_params = create_qaoa_variational_params(qaoa_circuit_params=self.circuit_params, params_type=self.circuit_properties.param_type,
    351                                                      init_type=self.circuit_properties.init_type, variational_params_dict=self.circuit_properties.variational_params_dict,
    352                                                      linear_ramp_time=self.circuit_properties.linear_ramp_time, q=self.circuit_properties.q, seed=self.circuit_properties.seed)

File c:\users\owenl\documents\openqaoa-main\openqaoa\qaoa_parameters\operators.py:610, in Hamiltonian.classical_hamiltonian(cls, terms, coeffs, constant)
    608         pauli_ops.append(PauliOp('Z', term))
    609     else:
--> 610         raise ValueError(
[...]

PR #75, opened 2022-07-22
added fix to knapsack and slackfreeknapsack random_instance

vishal-ph:
## Description

`Knapsack` and `SlackFreeKnapsack` problem classes convert a Knapsack problem into a QUBO problem solvable using QAOA. They also include a `random_instance(n_items)` method to generate a problem randomly. This method failed in the case of the randomly generated list of item weights being all equal. 

This was due to `weight_capacity` selected randomly between `np.min(weights), np.max(weights)`. However, in the all-equal case, this statement breaks. The fix takes care of this edge case.

- **Fixes [#48](/entropicalabs/openqaoa/issues/48)**

## Checklist

## Type of change

### DIFF
--- a/openqaoa/problems/problem.py
+++ b/openqaoa/problems/problem.py
@@ -642,7 +642,8 @@ class Knapsack(Problem):
 
         min_weights = np.min(weights)
         max_weights = np.max(weights)
-        if min_weights * n_items + 1 < max_weights * n_items :
+
+        if min_weights != max_weights:
             weight_capacity = np.random.randint(min_weights * n_items, max_weights * n_items)
         else:
             weight_capacity = np.random.randint(max_weights, max_weights * n_items)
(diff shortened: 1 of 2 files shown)


openqaoa/problems/problem.py
```diff
@@ -639,9 +639,15 @@ def random_instance(**kwargs):
 
         values = list(map(int, np.random.randint(1, n_items, size=n_items)))
         weights = list(map(int, np.random.randint(1, n_items, size=n_items)))
-        weight_capacity = np.random.randint(np.min(weights) * n_items, np.max(weights) * n_items)
-        penalty = 2 * np.max(values)
 
+        min_weights = np.min(weights)
+        max_weights = np.max(weights)
+        if min_weights * n_items + 1 < max_weights * n_items :
```
shahidee44, line 645: Since the edge case is when `min_weights == max_weights` ,maybe it makes sense to check for that instead. I'm not sure if im missing something here.

vishal-ph: fixed

Q-lds (approved):

PR #79, opened 2022-08-26
Setup pypi

Q-lds:
## Description
 - Organise setup.py into core packages, tests, and docs
- Fix security issue [nbconvert vulnerable to cross-site scripting (XSS) via multiple exploit paths](https://github.com/entropicalabs/openqaoa/security/dependabot/2)
- add pypi config files

## Checklist

- [ x] I have made corresponding updates to the documentation.

## Type of change

## How Has This Been Tested?


setup.py
```diff
@@ -1,42 +1,57 @@
...
 with open("README.md", "r") as fh:
     long_description = fh.read()
 
 requirements = [
     "sympy>=1.10.1",
-    "numpy==1.22.3",
-    "networkx==2.8",
+    "numpy>=1.22.3",
+    "networkx>=2.8",
     "scipy==1.8",
```
shahidee44, line 13: Hm should we also just do ">=" for the remaining packages or leave it as it is unless the other packages also throw incompatibility errors?

setup.py
```diff
@@ -1,42 +1,57 @@
...
     url="https://github.com/entropicalabs/openqaoa",
     description="A python SDK for Quantum Optimisation",
-    license='LICENSE.md',
+    license="LICENSE.md",
     long_description=long_description,
     long_description_content_type="text/markdown",
     classifiers=[
-        "Programming Language :: Python :: 3.8+",
-        "Operating System :: OS Independent"
+        "Programming Language :: Python :: 3.7",
```
shahidee44, line 46: I believe the documentation's main page has a statement regarding using python 3.8
Should that be corrected as well?
Maybe we can mention that the minimum python requirement is 3.7

setup.py
Q-lds: As far as I could test currently only  `scipy` requires the exact version. Have I forgotten to put the `>=` for any other?

setup.py
Q-lds: yes, that's a good point. I'll revert it to 3.8 for the moment. Otherwise, we can try the grid testing on github (i.e. running the test for 3.7, 3.8, 3.9)

Q-lds: Fixed

Q-lds:
@vishal-ph and @shahidee44 I think this is ready to be merged. We'll need to change some aspects of the pypi pipeline but that's ok I think for now (we'll need to update it to the real pypi index)

shahidee44 (approved):

PR #82, opened 2022-09-16
added plot_probabilities method in result class

EmilianoG-byte:
## Description

- Added a `plot_probabilities` method under the result class
- Nice and easy visualisation of the probabilities obtained for the basis states after the optimisation workflow.
- works for both state vector and shot based back-ends
- note: can add an additional argument to decide whether to plot states with zero probability or not.

## Checklist

## Type of change

## How Has This Been Tested?

Via UnitTest in the test_result_object.py file. Both for 'vectored' and 'qiskit.shot_simulator'.


openqaoa/optimizers/result.py
```diff
@@ -143,3 +143,54 @@ def plot_cost(self, figsize=(10, 8), label='Cost', linestyle='--', color='b', ax
...
+        if ax is None:
+            fig, ax = plt.subplots(figsize=figsize)
+
+        ax.bar(labels,probs, color=color)
+        ax.set_xlabel('Eigen-State')
+        ax.set_ylabel('Probability')
+        ax.set_title(label)
+        plt.xticks(rotation=70)
+        plt.grid(True, axis='y', linestyle='--')
+        plt.show
```
shahidee44, line 194: Should this line be `plt.show()` ?

shahidee44 (changes requested):

openqaoa/optimizers/result.py
```diff
@@ -143,3 +143,54 @@ def plot_cost(self, figsize=(10, 8), label='Cost', linestyle='--', color='b', ax
         ax.set_title('Cost history')
 
         return
+
+    def plot_probabilities(self, figsize = (10,8),label='Probability distribution',color='tab:blue', ax=None):
+
+        """
+        Helper function to plot the probabilities corresponding to each basis states (with prob != 0) obtained from the optimized result
```
shahidee44, line 150: We might want to consider having a warning in the docstring when using this method. It took quite some time for a graph from a 8-Qubit computation to be plotted.

Q-lds: Very good point! Two thoughts: 

1. If you want to plot something with more than, say, 8 qubits, then the plot contains the most probable states. The actual number should be a fraction that gets smaller as the number of qubits increases. Question: given that you have run a simulation over `25` qubits: how many states should we really want to plot? `2^25` is out of the question. But also, the most probable `1%` could potentially be very large.

2. This method should work also when plotting the results of a QPU with `O(100)` qubits, and the current implementation will just freeze your computer if you try to plot it. Now, let's assume we have 50 qubits. That is `2^50` possible states. You are doing shots, so in the worst case where each shot yields a different state, you only need to plot `#shots`. So, we could implement some general rules for displaying these states without freezing the machine

So, in both cases I think we should have clip the number of states we want to plot and let the user know that we did that and (perhaps) give them the opportunity to overwrite the warning and crash their laptops :D

openqaoa/optimizers/result.py
EmilianoG-byte: you are right! I just changed all the plt. lines to format only the ax object

openqaoa/optimizers/result.py
EmilianoG-byte: So, should I just choose an arbitrary number of states to show for all the cases or rather a dynamic one? If so, how to choose this number?

Q-lds: I think the main issue is not the number of data-points but the labels. So, I propose:
- We cap the number of labels in the plot to something like ~100
- We could include a panda's data frame result where you get get the statistics of larger distributions and where you can subsequently plot them on your own time and desire.

Thoughts, @vishal-ph @shahidee44  @EmilianoG-byte

vishal-ph: I like what the `qiskit.tools.visualization.plot_histogram` does here. As a user, you have an option to plot the most probable `x` states individually (with their labels displayed on the x-axis). The remaining state probabilities are added in a single bar at the end (with a label, for instance, "remaining states/ other states"). 
Therefore, in total the plotter will only show `x+1` bars on the picture.

shahidee44: I think even if we just limit the number of labels that can be the, just trying to plot 2^N bars will not scale very well in the long run. So we would still have to reduce the number of data-points here.

Since we are thinking slightly long term here, I'd say we should have some kind of plot cap. (Maybe "best" 100 states ranked by probability(?)). 

At the same time, I don't think we should overcomplicate some of these methods since we intend to add more methods that could visualize the outputs in various different ways. The second point you propose would probably make sense as a seperate method within the Result object.

Alternatively, we can just produce some kind of error if the user tries to plot to many data-points. Like raising a warning if they try to plot more than 100 data-points.

openqaoa/optimizers/result.py
```diff
@@ -143,3 +143,53 @@ def plot_cost(self, figsize=(10, 8), label='Cost', linestyle='--', color='b', ax
...
+
+        if ax is None:
+            fig, ax = plt.subplots(figsize=figsize)
+
+        ax.bar(labels,probs, color=color)
+        ax.set_xlabel('Eigen-State')
+        ax.set_ylabel('Probability')
+        ax.set_title(label)
+        ax.tick_params(axis='x', labelrotation = 70)
+        ax.grid(True, axis='y', linestyle='--')
```
shahidee44, line 193: Hmm. When I run this method in a jupyter notebook, the plot doesnt show until the user does `matplotlib.pyplot.show()` . Is this the intended behaviour?

openqaoa/optimizers/result.py
Q-lds: Ok, then I think we can go with what Vishal proposed. 

Sounds fair, @EmilianoG-byte ?

EmilianoG-byte: Ok. So, if I understood correctly, we will give the option to the user to select the top `x` states, and set a default value for `x` (and equal upper bound) of 100 states. Is that correct?

openqaoa/optimizers/result.py
EmilianoG-byte: it works well for me, could you tell me more about the instance you tried it on?

Screenshot 2022-09-27 at 23 02 57
Screenshot 2022-09-27 at 23 02 57 2

openqaoa/optimizers/result.py
shahidee44: Yes. Then we just sum up the probabilities of the remaining state and add that as a final bar in the graph. Therefore, we would have `x+1` bars in the bar graph. (Maybe you can try to play around with the default value and see whats a good default to set such that the graph is still easily readable too)

EmilianoG-byte: Is it okay if I also increase the figsize to include more states? So far I was using the same as for the `plot_cost` method which is (10,8). If I do change mine, should I change the one for the cost as well?

openqaoa/optimizers/result.py
EmilianoG-byte: I have done the changes, let me know what you think :)

openqaoa/optimizers/result.py
EmilianoG-byte: have done the changes, let me know what you think :)

openqaoa/optimizers/result.py
Q-lds: Can we check your two versions of mathplotlib?

openqaoa/optimizers/result.py
```diff
@@ -156,6 +157,90 @@ def plot_cost(
...
+        total = len(states)
+
+        # number of states that fit without distortion in figure
+        upper_bound = 40
+        # default fontsize
+        font = 'medium'
+
+        if states_to_keep:
+            if states_to_keep > total:
+                raise TypeError('states_to_keep must be smaller than the total number of states')
```
Q-lds, line 208: I think it needs to be a ValueError, and it could be useful to have a reference of what the total is

replace to `raise ValueError(f'states_to_keep must be smaller or equal to the total number of states {total}')`

And should it say `smaller or equal`, right?

openqaoa/optimizers/result.py
```diff
@@ -156,6 +157,90 @@ def plot_cost(
...
+
+        # total number of states / number of states with != 0 counts for shot simulators
+        total = len(states)
+
+        # number of states that fit without distortion in figure
+        upper_bound = 40
+        # default fontsize
+        font = 'medium'
+
+        if states_to_keep:
```
Q-lds, line 206: Since this is an `int` should we call `n_states_to_keep`? Too pedantic?

openqaoa/optimizers/result.py
Q-lds: I think it is quite nice! I have just added a few comments.

Also, note that you may need to pull the branch, since I have made a few changes to make sure the tests were all green on our end :)

openqaoa/optimizers/result.py
EmilianoG-byte: I'm using '3.5.3'

openqaoa/optimizers/result.py
EmilianoG-byte: I like it, specially to keep it coherent witht the rest of the code. I will change it then :)

openqaoa/optimizers/result.py
EmilianoG-byte: Thank you for the comments! I implemented all of them, the only difference is that I included 'number of states in measurement outcome' in the exception. This is to make it clear because sometimes the total states differ from 2^qubits when doing shots simulators and some states have zero counts

openqaoa/optimizers/result.py
EmilianoG-byte: Is anybody else having the error `Pauli terms should be of type PauliOp and not <class 'openqaoa.qaoa_parameters.operators.PauliOp'>` when compiling the problem? This was not happening before

Q-lds: Not really, I have tried, and all seems fine. Can you let us know how to reproduce the error?

Q-lds:
Ok! Besides the potential issue regarding the Pauli op error, I think this PR is ready to be merged! (cc @vishal-ph )

openqaoa/optimizers/result.py
vishal-ph: > Ok! Besides the potential issue regarding the Pauli op error, I think this PR is ready to be merged! (cc @vishal-ph )

I see, there are still some changes requested by Shahidee. Are those already implemented? If yes, lets resolve the comments and we can proceed with merge

vishal-ph:
> Ok! Besides the potential issue regarding the Pauli op error, I think this PR is ready to be merged! (cc @vishal-ph )

I see, there are still some changes requested by Shahidee. Are those already implemented? If yes, lets resolve the comments and we can proceed with merge

openqaoa/optimizers/result.py
EmilianoG-byte: Initially, I got it when compiling a QUBO for ShortPath Problem, but the problem is there for something as simple as:


`problem_qubo = QUBO(3, [[0],[1],[2]], [1,1,1])`
`q = QAOA()`
`q.compile(problem_qubo)`
`q.optimize()`

EmilianoG-byte: Ok, I just reloaded VS code and everything works fine. I don't what was going on. Sorry for the confusion :/

EmilianoG-byte:
>

EmilianoG-byte:
> > Ok! Besides the potential issue regarding the Pauli op error, I think this PR is ready to be merged! (cc @vishal-ph )
> 
> I see, there are still some changes requested by Shahidee. Are those already implemented? If yes, lets resolve the comments and we can proceed with merge

You man regarding the use of plt.show(), right? With my version of Matplotlib it is not necessary. I am unsure what version is Shahidee using

openqaoa/optimizers/result.py
shahidee44: I'm using `3.4.3`

But I think i figured it out. It seems that I wasn't running `%matplotlib inline` in my notebook.
If that magic method isnt used, the user has to specifically request `pyplot` to show the plot with `matplotlib.pyplot.show()`.

vishal-ph: perhaps you won't need if you include `plt.show()` in the body of the function, no?

shahidee44: Yes. As long as `plt.show()` is ran within the function, we wont need to have the user to run it by himself.

EmilianoG-byte: I see. I will add it then! Should I also add it to the `plot_cost` method?

vishal-ph (approved):

openqaoa/optimizers/result.py
shahidee44: Looks like it might just be some versioning difference where this is an issue. We've decided to just integrate this version of the PR without the implementation of `plt.show()` :)

PR #88, opened 2022-09-23
fixed varational params plot method

raulconchello:
Fixed the .plot method for the varational params class


vishal-ph:
Hi, let's add some simple unit tests for these plot functions. For e.g. if some future version of matplotlib makes some functionality obsolete, we don't want them to fail silently. A test for example, that checks whether the plot method works and generates a figure would catch if something gets deprecated in future.

vishal-ph (approved):

PR #98, opened 2022-10-06
Dev parallelization

shaohenc:
Addition of pyQuil-specific parallelization passes for QC runs:
1. Unfencing pass : allows 2Q gates to be executed in parallel by removing global fence statements (which are by default enabled)
2. trivial_parallelization pass: re-order pyQuil program according to 'trivial' commutation rules, i.e. re-arranging order of gates that act on independent sets of qubits.
3. [WIP] commutation_parallelization pass: re-order pyQuil program according to Pauli matrix commutation rules.


vishal-ph:
@shaohenc, what is the status of this PR? If this work is near completion, I suggest we clean it up and add the feature to the active the branches

PR #104, opened 2022-10-20
Dev bipartite graphs

vishal-ph:
## Description

- Add a tutorial notebook solving MVC with OpenQAOA 

## Type of change

- [x] New tutorial notebook


Q-lds:
A few ideas to improve this PR:
- The notebook contains a lot of functions: shouldn't these functions be part of a regular `.py` file? Maybe some utilities file in the problem folder?
- We need tests for the new functions
- The JSON inputs should be moved somewhere else: perhaps in a new folder, `examples/problem_input`?
- The notebook itself should be moved to the example section. It could also be expanded a bit more

Q-lds (approved):

Issue #105, opened 2022-10-20
Unintended behaviour/error-catching bug when using workflows with IBM backend


shaohenc:
### Prerequisites
Before raising this issue, I have already checked that I am:

- [x] running the latest version
- [x] made sure that this issue has not already been filed

### Describe the bug
Using workflows, if the QUBO problem to be compiled contains more qubits than the device chosen, running `q.optimise()` will just pass, without returning an error message.

### To Reproduce
1. Choose an IBM device (eg: `ibm_nairobi` with 6 qubits)
2. Create a QUBO problem requiring more qubits than the device (eg : a 7-qubit MaxCut problem)
3. Run OQ as usual, and finally run `q.optimise()`
4. Cell passes, but it should have failed.

### Expected behavior 
An error message should be thrown to inform the user that the problem and device are incompatible.

vishal-ph:
PR 108 fixes this issue

PR #108, opened 2022-10-25
Bugfix: QPU qubit overflow

shahidee44:
## Description

- Added `n_qubits` attribute to `DeviceQiskit` that extracts the maximum number of qubits compatible for the particular IBMQ Backend selected
- Added qubit number check between `CircuitParams` Object and `DeviceQiskit`. An Exception is raised if the number of qubits in the backend is less than the number of qubits required to create the circuit.
- Updated unittests in test_qiskit_qpu.py

- Added `n_qubits` attribute to `DeviceAWS` that extracts the maximum number of qubits compatible for the particular Braket Backend selected. (Since not all backends may have the total number of qubits available, a function that catches the possibility that the number of qubits is not available is added for this Device object.)
- Added qubit number check between `CircuitParams` Object and `DeviceAWS`. Exception is raised as above.
- Updated unittests in test_braket_qpu.py

- Added `n_qubits` attribute to `DevicePyQuil` that extracts the maximum number of qubits compatible for the particular PyQuil QPU Backend selected. 
- Added qubit number check between `CircuitParams` Object and `DevicePyquil`. Exception is raised as above.

- The above resolves PR #105 

- Updated Exception catching in `OptimizeVQA` should help prevent silent crashes.
- Updated unittests to check for silent errors.
- Minor changes to `OptimizeVQA`s `optimize` method. (Return is moved outside of the try/except/finally loop. If it is within the finally tab, the exception we want is not raised.) This resolves PR #132 
- Renamed `test_optimisers.py` to `test_optimizers.py`

## Checklist

## Type of change

## How Has This Been Tested?

test_all.py


shahidee44:
Comments on PyQuil Backend:

The backend does not fail silently like the others when a larger circuit is specified than is supported by the actual QPU.
The main reason for this is that within the init function of the PyQuilQPUBackend Class, the paramteric circuit is compiled for the particular QPU selected. This compilation step throws the error. Since the error is not thrown within the get_counts method, the code does not silently fail.
The new addition to the code just makes sure that the failure due to "qubit overflow" has a consistent message throughout all the backends.

Additional Comments:
With the slightly different design of the `DevicePyQuil`, `n_qubits` attribute assigned at the initialisation of the class since this information is readily available. This is not True for `DeviceQiskit` and `DeviceAWS` where the `check_connection()` must first be executed before `n_qubits` becomes available.
There is no direct effect in the code when `n_qubits` is assigned. Though we may want to consider if it might be appropriate to have similar behaviours for all Device Objects.

openqaoa/backends/qpus/qaoa_braket_qpu.py
```diff
@@ -92,7 +92,10 @@ def __init__(self,
         else:
             raise Exception('Error connecting to AWS.')
             
-        self.parametric_circuit = self.parametric_qaoa_circuit()
+        if self.device.n_qubits < self.n_qubits:
+            raise Exception('There are lesser qubits on the device than the number of qubits required for the circuit.')
+            
+        self.parametric_circuit = self.parametric_qaoa_circuit
```
vishal-ph, line 98: self.parametric_qaoa_circuit with parentheses or not?

tests/test_qpu_braket.py
```diff
@@ -360,37 +362,56 @@ def test_exceptions_in_init(self):
         aws_device = DeviceAWS('arn:aws:braket:::device/quantum-simulator/amazon/sv1')
         
         QAOAAWSQPUBackend(circuit_params, aws_device, shots, None, None, True, 1.)
-     
+    
+    @pytest.mark.qpu            
+    def test_remote_qubit_overflow(self):
```
vishal-ph, line 367: I dont fully understand this test? How would it overflow? SV1 should be able to run 10 qubits, no?

vishal-ph: Oh I see, you are basically creating a 100 qubit cost Hamiltonian, and expecting that for SV1, AWS will output an overflow error? Did I understand it correctly?

vishal-ph (approved):

PR #109, opened 2022-10-26
RQAOA workflow

raulconchello:
# Description
The RQAOA workflow class has been refactored. 
Now it is using QAOA objects to solve the qaoa steps, and those are stored. Therefore, all features of QAOA objects are now accepted by the RQAOA objects.

Some test have been written.

## TO DO 
- [x] Change the Notebook example for RQAOA
- [x] Check and change docs
- [x] Documentation
- [x] Improve RQAOAResults class
- [x] Tests for RQAOAResults class


raulconchello:
## QUESTIONS
- [ ] Should rqaoa_type ('adaptive', 'custom') exist? It could be determined auto, and we could mix 'adaptive' and 'custom' -e.g. schedule=[10,10, Auto, Auto, ...]-.
- [ ] Should we test features of QAOA when testing the RQAOA? E.g. test different variational parameters.
- [ ] Move the rqaoa.py file?
- [ ] Are the results stored ok in qaoa.results (`dict`)?

openqaoa/workflows/optimizer.py
```diff
@@ -430,42 +458,149 @@ def compile(self, problem: QUBO = None, verbose: bool = True):
...
+
+        # create a different max_terms function for each type 
+        if self.rqaoa_parameters.rqaoa_type.lower() == "adaptive":
+            f_max_terms = rqaoa.ada_max_terms  
         else:
-            raise f'rqaoa_type {self.rqaoa_parameters.rqaoa_type} is not supported. Please selet either "adaptive" or "custom'
+            f_max_terms = rqaoa.max_terms #? rqaoa.max_terms -> rqaoa.custom_max_terms ?
+
+        # If above cutoff, loop quantumly, else classically
+        while n_qubits > n_cutoff:
```
Q-lds, line 521: I believe it could make sense to have a result object for this part of the algorithm and make sure that anything of interest in this while-loop is logged.

Probably the way to go is to have a logger here and then have something that formats the result to be presented to the user.

 Pinging @shahidee44, as he may have some good comments for this feature

openqaoa/workflows/optimizer.py
```diff
@@ -430,42 +458,149 @@ def compile(self, problem: QUBO = None, verbose: bool = True):
...
+            counter += 1
+
+        # Solve the new problem classically
+        cl_energy, cl_ground_states = ground_state_hamiltonian(problem.hamiltonian)
+
+        # Retrieve full solutions including eliminated spins and their energies
+        full_solutions = rqaoa.final_solution(
+            self.elimination_tracker, cl_ground_states, self.problem.hamiltonian)
+
+        # Compute description dictionary containing all the information
```
Q-lds, line 559: Answering a question you posed: I think it make sense to have these things formatted nicely and to create a rqaoa-result object (mimicking the structure of the QAOA one)

raulconchello: I have created a RQAOAResults class to do so in my last commit, but for now is very different to the QAOA Results, I will change it.

But I am not sure where this class should be, for now is in optimizer.py, since the actual result.py file is in the classical optimizers methods.

raulconchello: In this object, I have included a method: `get_qaoa_steps`, which returns a list of the qaoa objects that have been used in each step of the RQAOA

And also `get_elimination_rules`, which returns a list that keeps track of each elimination that has occurred.

openqaoa/workflows/optimizer.py
raulconchello: I think the RQAOAResults object has all the information to keep track of what has happened in each step (see the conversation below). I don't know if that is good enough, or if we should have a different object for that.

shahidee44: Raul's RQAOAResults object is quite comprehensive. It might be worth reviewing what he has added. He's showed me some of the features that he has incooporated it and I believe its sufficient for our use-case. 
Unless you have anything else specific in mind, I think it looks good and is quite functional.

openqaoa/workflows/optimizer.py
raulconchello: I have changed again the RQAOAResults class. I think now it makes more sense. The class inherits from `dict`, therefore `RQAOA.results` is a dictionary itself, but we can also create all the methods that we want. E.g. `RQAOAResults .get_hamiltonian_step(I)`, which returns the hamiltonian of the i-th intermediate QAOA step. 

```
r = RQAOA()
r.complie(problem)
r.optimize()
```
```
r.results
```
> {'solution': {'001011': -7.0, '110100': -7.0},
>  'classical_output': {'minimum_energy': -4.0,
>   'optimal_states': ['000', '111']},
>  'elimination_rules': [{(0, 2): -1.0}, {(1, 3): -1.0}, {(0, 3): -1.0}],
>  'schedule': [1, 1, 1],
>  'intermediate_steps': [{'QUBO': <openqaoa.problems.problem.QUBO at 0x282d0d93520>,
>    'QAOA': <openqaoa.workflows.optimizer.QAOA at 0x282d0d93130>},
>   {'QUBO': <openqaoa.problems.problem.QUBO at 0x282d160e070>,
>    'QAOA': <openqaoa.workflows.optimizer.QAOA at 0x282d15c6310>},
>   {'QUBO': <openqaoa.problems.problem.QUBO at 0x282d20a31c0>,
>    'QAOA': <openqaoa.workflows.optimizer.QAOA at 0x282d0c09c70>}],
>  'number_steps': 3}
```
r.results['solution']
```
>{'001011': -7.0, '110100': -7.0}
```
r.results.get_hamiltonian_step(2)
```
><openqaoa.qaoa_parameters.operators.Hamiltonian at 0x282d139bfa0>

Also, I have erased all the getters that were trivial. And now `r.results['intermediate_steps' ]` is a list of the intermediate QUBO and QAOA.

Let me know what you think.

Q-lds: I like it!

My only comment is that I would want to have a JSON serialisable version of it already built in.

This current result can (perhaps) be saved with `pickle` but I am not 100% sure about it.

Either way, it would be good to have something like `r.results.dump()`, which:
- converts the object into a json
- converts the QUBO and the QAOA object into json too

Probably, the second step should be done elsewhere. That is, when you do `r.result.dump()`, then the `QUBO.dump()` and `QAOA.dump()` are called

openqaoa/workflows/optimizer.py
Q-lds: Marking this as resolved

openqaoa/workflows/optimizer.py
```diff
@@ -281,6 +181,149 @@ def set_classical_optimizer(self, **kwargs):
         self.classical_optimizer = ClassicalOptimizer(**kwargs)
         return None
 
+    def compile():
+        raise NotImplementedError
+
+    def optimize():
+        raise NotImplementedError
+
```
Q-lds, line 189: Why so many spaces?

openqaoa/workflows/optimizer.py
```diff
@@ -430,42 +536,206 @@ def compile(self, problem: QUBO = None, verbose: bool = True):
...
+        self._q.classical_optimizer = self.classical_optimizer
 
-    def optimize(self):
+        # compile qaoa object
+        self._q.compile(problem, verbose=verbose)
+
+
+        if self.rqaoa_parameters.rqaoa_type.lower() == "custom":
+
+            n_cutoff = self.rqaoa_parameters.n_cutoff
```
Q-lds, line 554: I wouldn't introduce new variables but use the self. @vishal-ph ?

openqaoa/workflows/optimizer.py
```diff
@@ -430,42 +536,206 @@ def compile(self, problem: QUBO = None, verbose: bool = True):
...
+            spin_map = rqaoa.spin_mapping(problem, max_terms_and_stats)
+            # Eliminate spins and redefine problem
+            new_problem, spin_map = rqaoa.redefine_problem(problem, spin_map)
+
+            # Extract final set of eliminations with correct dependencies and update tracker
+            eliminations = {(spin_map[spin][1],spin):spin_map[spin][0] for spin in sorted(spin_map.keys()) if spin != spin_map[spin][1]}
+            elimination_tracker.append(eliminations)
+
+            # Extract new number of qubits
+            n_qubits = new_problem.n
```
Q-lds, line 622: Here you overwrite the variable defined in line `593`. I'd use a different variable name instead

openqaoa/workflows/optimizer.py
```diff
@@ -430,42 +536,206 @@ def compile(self, problem: QUBO = None, verbose: bool = True):
...
+            
+            # Ensure there are enough steps in the schedule
+            assert np.abs(n_qubits - n_cutoff - counter) <= sum(self.rqaoa_parameters.steps),\
+                f"""Schedule is incomplete, add 
+                    {np.abs(n_qubits - n_cutoff - counter) - sum(self.rqaoa_parameters.steps)} 
+                    more eliminations"""
+
+            self.compiled = True
+        
+        elif not self.rqaoa_parameters.rqaoa_type.lower() == "adaptive":
```
Q-lds, line 570: I am not sure I get this:

if custom --> do something
elif not adaptive -- > raise an execption?

Is this correct?

raulconchello: Yes, so 
if not custom nor adaptative -> raise exception 
if custom -> do something

Q-lds: Maybe change the exception clause to `if not self.rqaoa_parameters.rqaoa_type.lower() in [''custom', 'adaptive']`

I don't think the elif is the correct clause for this scenario

openqaoa/workflows/optimizer.py
raulconchello: The `n_qubits` variable? but I need to change it to update the number of qubits with the new problem

openqaoa/workflows/optimizer.py
raulconchello: I added them to make the code more readable but if it's not optimal I'll use the self. variables directly

openqaoa/workflows/optimizer.py
raulconchello: I was trying to do that by first converting all attributes to a dictionary by using the `convert2serialize` function but I get an error:
```
RecursionError: maximum recursion depth exceeded in comparison
```
It comes from when trying to convert the QAOA.backend: `convert2serialize(QAOA.backend)`

Should I use this function or create another one that is not recursive?

Q-lds: Yes. And I don't think that one can be serialised easily (and probably it shouldn't).

Ok, let's not do too many things all at once. Let's first close this PR and then move on and make first the important parts of the QAOA object serialisable, and then let's deal with the RQAOA. What do you say?

openqaoa/workflows/optimizer.py
Q-lds: Yeah, I know. I think I prefer shorter scripts. Let's leave it like this for the moment  then :)

openqaoa/workflows/optimizer.py
raulconchello: Okay, yes, that makes sense!

So, do I implement the r.result.dump() method but leaving the QAOA and QUBO objects not converted or I don't implement the method at all, for now?

Q-lds: Leave it like this at the moment, and let's do a another PR right after this one to fix the formatting. It is an important feature to have!

docs/source/rqaoa.rst
```diff
@@ -14,14 +14,14 @@ The development of this method is associated with an internal research project a
 
 Custom RQAOA
 ------------
-.. autoclass:: openqaoa.rqaoa.custom_rqaoa
+.. autoclass:: openqaoa.workflows.optimizer.RQAOA
```
vishal-ph, line 17: Need to review whether it makes sense to do the distinction now. Its the same object now

raulconchello: Yes, I think it doesn't make sense to have the distinction now

raulconchello: And I don't know if it should be here or only in workflows.rst

vishal-ph: true that! Probably not since this function is not present in this file anymore

Q-lds: Since the PR originates from outside OQ I don't think I can build the docs ... but they will be automatically built once we merge the PR to dev.

Then we can fix the documentation (if needs be)

Q-lds (approved):

PR #114, opened 2022-10-28
Refactor the problem class

Jaakik:
## Description

Refactoring Problem 

## Checklist
- [x] generating random instances without kwargs 
- [x] Encoding of problem classes


openqaoa/problems/helper_functions.py
```diff
@@ -29,41 +29,34 @@ def convert2serialize(obj):
...
-            Keyword arguments where keys are supposed to be the expected params
-
-    Returns:
-        A tuple with the actual expected parameters if they are found in kwargs.
-
-    Raises:
-        ValueError: 
-            If one of the expected arguments is not found in kwargs and its 
-            default value is not specified.
+    Converts the weights from a [0, 1] encoding to an Ising problem [-1, 1] 0 is mapped to +1 and 1 to -1 respectively
```
shahidee44, line 34: Is 0 mapped to +1 or -1. I thought it should be mapped to -1.

openqaoa/problems/helper_functions.py
```diff
@@ -29,41 +29,34 @@ def convert2serialize(obj):
...
+
+    Parameters
+    ----------
+    terms: list[list]
```
shahidee44, line 38: Is this typing correct. Should its be `list[list]` or `list[str]`

openqaoa/problems/helper_functions.py
```diff
@@ -29,41 +29,34 @@ def convert2serialize(obj):
...
+        terms of the hamiltonian
+    weights: list[float]
+
+    Returns
+    -------
+    terms_weights: tuple(list[list],list[float])
```
shahidee44, line 44: Same as above.

openqaoa/problems/problem.py
```diff
@@ -234,32 +243,23 @@ def coordinates(self, input_coordinates):
...
         Returns
         -------
             A random instance of the Traveling Salesman problem.
         """
-        n_cities = check_kwargs(["n_cities"], [None], **kwargs)[0]
-        seed = kwargs.get("seed")
-
-        if isinstance(seed, int):
-            np.random.seed(seed)
+        np.random.seed(SEED)
```
shahidee44, line 261: Do you know if doing the seed this way might cause problems. i.e. if the user wants to set his own numpy seed before running this function. Also does repeatedly running random_instance create the same instance all the time. Looks like with this setup it might.

openqaoa/problems/problem.py
```diff
@@ -763,49 +736,18 @@ def get_qubo_problem(self):
...
-
-        for variable, linear_term in enumerate(linear_terms):
-            ising_terms.append([variable])
-            ising_coeffs.append(linear_term)
-
-        ising_terms.append([])
-        ising_coeffs.append(constant_term)
-
-        return QUBO(n_variables, ising_terms, ising_coeffs)
+        return QUBO(n_variables, ising_terms, weights)
```
shahidee44, line 742: is this supposed to be `QUBO(n_variables, ising_terms, ising_weights)` ?

openqaoa/problems/problem.py
```diff
@@ -14,35 +14,44 @@
...
 from openqaoa.qaoa_parameters.operators import Hamiltonian
 
+SEED = 1234
+
 
 class Problem(ABC):
     @staticmethod
     @abstractmethod
-    def random_instance(**kwargs):
+    def random_instance(problem_size):
```
shahidee44, line 32: We should add an explanation of this input parameter somewhere in the docstring.
Probably in every class since problem size in each problem class would/could mean different things.

openqaoa/problems/problem.py
```diff
@@ -1144,34 +1037,24 @@ def __init__(self, G, source, dest):
...
-        n_nodes, edge_probability, seed, source, dest = check_kwargs(
-            ["n_nodes", "edge_probability", "seed", "source", "dest"],
-            [None, None, 1234, 0, 1],
-            **kwargs,
-        )
+        np.random.seed(SEED)
+        n_nodes = np.random.randint(1, problem_size)
+        edge_probability = np.random.uniform(0, 1)
+        source = np.random.randint(0, n_nodes)
+        dest = np.random.randint(0, n_nodes)
```
shahidee44, line 1055: If this function generates a source and dest node of the same index, it will error out since the ShortestPath class requires the nodes to have a different index.

openqaoa/problems/problem.py
```diff
@@ -1144,34 +1037,24 @@ def __init__(self, G, source, dest):
...
         A random instance of the Shortest Path problem.
         """
-
-        n_nodes, edge_probability, seed, source, dest = check_kwargs(
-            ["n_nodes", "edge_probability", "seed", "source", "dest"],
-            [None, None, 1234, 0, 1],
-            **kwargs,
-        )
+        np.random.seed(SEED)
+        n_nodes = np.random.randint(1, problem_size)
```
shahidee44, line 1052: Shortest path should have at least 2 nodes. This would allow for just 1 node, which doesnt work.

openqaoa/problems/problem.py
```diff
@@ -829,24 +771,17 @@ def __init__(self, values, weights, weight_capacity, penalty):
...
         -------
             A random instance of the Knapsack problem.
         """
-        n_items = check_kwargs(["n_items"], [None], **kwargs)[0]
-        seed = kwargs.get("seed")
-
-        if isinstance(seed, int):
-            np.random.seed(seed)
+        np.random.seed(SEED)
+        n_items = np.random.randint(1, problem_size)
```
shahidee44, line 784: Minimum size of this problem should be more than 1.

openqaoa/problems/problem.py
```diff
@@ -641,24 +621,17 @@ def penalty(self, input_penalty):
...
         -------
             A random instance of the Knapsack problem.
         """
-        n_items = check_kwargs(["n_items"], [None], **kwargs)[0]
-        seed = kwargs.get("seed")
 
-        if isinstance(seed, int):
-            np.random.seed(seed)
+        np.random.seed(SEED)
+        n_items = np.random.randint(1, problem_size)
```
shahidee44, line 634: Minimum size of this problem should be more than 1.

openqaoa/problems/problem.py
```diff
@@ -491,32 +482,21 @@ def G(self, input_networkx_graph):
...
         Returns
         -------
             A random instance of the Maximum Cut problem.
         """
-        n_nodes, edge_probability = check_kwargs(
-            ["n_nodes", "edge_probability"], [None, None], **kwargs
-        )
-        seed = kwargs.get("seed", None)
+        np.random.seed(SEED)
+        n_nodes = np.random.randint(1, problem_size)
```
shahidee44, line 495: Minimum size of this problem should be more than 1.

shahidee44 (changes requested):

openqaoa/problems/helper_functions.py
Jaakik: @vishal-ph ?

openqaoa/problems/helper_functions.py
Jaakik: I think it's already correct this way, @vishal-ph ?

openqaoa/problems/problem.py
Jaakik: Yes !

openqaoa/problems/problem.py
Jaakik: Yes, I have pointed this out with @vishal-ph and apparently they want this behaviour for reproducibility.

Jaakik:
I solved these issues and refactor some minor things, but I think the user should specify his seed in the workflow or we generate a random seed for each instance of the problem class. @shahidee44  @vishal-ph

openqaoa/problems/problem.py
```diff
@@ -615,12 +590,12 @@ def random_instance(**kwargs):
         -------
             A random instance of the Number Partitioning problem.
         """
-        n_numbers = check_kwargs(["n_numbers"], [None], **kwargs)
+        if problem_size < 2:
```
shahidee44, line 593: Shouldnt this be `<= 2` if the problem size should be at least 2.

openqaoa/problems/problem.py
```diff
@@ -1306,34 +1213,25 @@ def __init__(self, G, source, dest):
...
         -------
         A random instance of the Shortest Path problem.
         """
 
-        n_nodes, edge_probability, seed, source, dest = check_kwargs(
-            ["n_nodes", "edge_probability", "seed", "source", "dest"],
-            [None, None, 1234, 0, 1],
-            **kwargs,
-        )
+        if problem_size < 2:
```
shahidee44, line 1228: Same as above

PR #125, opened 2022-11-07
Bugfix: Running .optimize() for a RQAOA object not working as expected

raulconchello:
Before, if one optimized twice a RQAOA object, the second time wasn't working as expected. Because the self._q.was compiled with the last problem of the first optimization.  

## TODO
- [x] Every time that .optimize() is run the whole RQAOA process should restart
- [x] Check that now everything works as expected
- [x] Write some test that can check that this is the case

### DIFF
--- a/openqaoa/workflows/optimizer.py
+++ b/openqaoa/workflows/optimizer.py
@@ -684,6 +684,9 @@ class RQAOA(Optimizer):
         n_qubits = problem.n
         counter = self.rqaoa_parameters.counter 
 
+        # initialize the results object
+        self.results = RQAOAResults()
+
         # create a different max_terms function for each type 
         if self.rqaoa_parameters.rqaoa_type == "adaptive":
             f_max_terms = rqaoa.ada_max_terms  
@@ -692,6 +695,9 @@ class RQAOA(Optimizer):
 
         # If above cutoff, loop quantumly, else classically
         while n_qubits > n_cutoff:
+
+            # Compile qaoa with the problem
+            self._q.compile(problem, verbose=False)
 
             # Run QAOA
             self._q.optimize()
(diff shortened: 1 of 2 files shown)


tests/test_workflows.py
```diff
@@ -655,6 +655,17 @@ def _run_rqaoa(self, type, problem, n_cutoff=5, eliminations=1, p=1, param_type=
 
         return r.results.get_solution()
 
+    def test_rqaoa_complie_multiple_times(self):
```
vishal-ph, line 658: *compile typo

openqaoa/workflows/optimizer.py
```diff
@@ -682,7 +682,10 @@ def optimize(self, verbose=False):
         problem = self.problem  
         n_cutoff = self.rqaoa_parameters.n_cutoff
         n_qubits = problem.n
-        counter = self.rqaoa_parameters.counter 
+        counter = self.rqaoa_parameters.counter
+
+        # copy the original qaoa object
+        q = copy.deepcopy(self._q)
```
vishal-ph, line 688: perhaps it makes sense to use a more descriptive name for this? `q` does not tell me that this is the QAOA with the original problem.

tests/test_workflows.py
```diff
@@ -655,6 +655,17 @@ def _run_rqaoa(self, type, problem, n_cutoff=5, eliminations=1, p=1, param_type=
 
         return r.results.get_solution()
 
+    def test_rqaoa_compile_multiple_times(self):
```
vishal-ph, line 658: should it be `test_rqaoa_optimize_multiple_times`?

openqaoa/workflows/optimizer.py
raulconchello: but this q is going to change at each RQAOA step

tests/test_workflows.py
raulconchello: true

openqaoa/workflows/optimizer.py
raulconchello: Exactly

vishal-ph (approved):

PR #144, opened 2022-12-05
Changing the asdict and dumps methods for the workflows classes and adding experiment tags and uuids

raulconchello:
This PR modifies `asdict` and `dump(s)` workflows methods.
Also, it implements `set_exp_tags` and `set_header` methods for both qaoa and rqaoa.
Also, implements the generation of job ids for simulators.

### Identification
The attribute `.header` has been implemented for the workflows (qaoa and rqaoa). `(r)qaoa.id` is the following dictionary:
```
(r)qaoa.id ={
     'atomic_uuid': 'a9a34e30-a1ea-4208-8352-d9bd075113ed',
     'experiment_uuid': '5d3db7f9-55c0-47e5-97d6-d509d290be2e',
     'project_uuid': '8353185c-b175-4eda-9628-b4e58cb0e41b',
     'algorithm': 'qaoa',
     'name': 'test',
     'run_by': 'raul',
     'provider': '-',
     'target': '-',
     'cloud': 'local',
     'client': '-',
     'qubit_number': 9,
     'qubit_routing': '-',
     'error_mitigation': '-',
     'error_correction': '-',
     'execution_time_start': 1672989162,
     'execution_time_end': 1672989163,
     'metadata': {'tag1': 'value1', 'tag2': 'value2'}
}
``` 

 The user is responsible to set `project_uuid, name, run_by, provider, target, cloud, client, qubit_routing, error_mitigation, error_correction` using the `set_header` method:
```
q = QAOA()
q.set_header(
    project_uuid="8353185c-b175-4eda-9628-b4e58cb0e41b", 
    name="test", 
    run_by="raul", 
    provider="-", 
    target="-", 
    cloud="local", 
    client="-", 
    qubit_routing="-", 
    error_mitigation="-", 
    error_correction="-"
    )
```

### Tags
The attribute `.exp_tags` has also been implemented. It is also a dictionary. It is initialized as an empty dictionary. The user can fill it with any hyperparameter or any other tag using the `.set_exp_tags` method, as:
```
q = QAOA()
q.set_exp_tags(tags={'tag1': value1, 'tag2': value2})
```

### Results `.asdict` method
The `.asdict` method has been implemented for both qaoa and rqaoa results:
```
q = QAOA()
q.compile(problem)
q.optimize()
results_asdict = q.results.asdict()
```
This method returns a serializable dictionary containing all the information contained in the results object.

##### `.asdict`
This method returns a serializable dictionary containing all the information contained in the qaoa (or rqaoa) object. It also includes the results data.
```
q = QAOA()
q.compile(problem)
q.optimize()
q_asdict = q.asdict()
```
The dictionary has the following structure:

```
{
   'header': {...},
   'data': {
        'exp_tags': {...},
        'input_problem': {...},
        'input_parameters': {...},
        'results': {...}
   }
}
``` 
- `'header'` includes the `(r)qaoa.header` dictionary + `'metadata'`.
- `'exp_tags`' includes all the tags in the `(r)qaoa.exp_tags` dictionary. 
- `'input_problem'` is the serialization of the problem that we want to solve using the workflow.
- `'input_parameters'` includes all the parameters set through the `.set_device`, `.set_circuit_properties`, `.set_backend_properties`, `.set_classical_optimizer` and `.set_rqaoa_parameters` methods.
- `'results'` is filled using `(r)qaoa.result.asdict()`.

##### `.dumps`
This method returns a json string containing the dictionary `(r)qaoa.asdict()`:
```
q = QAOA()
q.compile(problem)
q.optimize()
q_json = q.dumps(indent=1)
```

##### `.dump`
This method creates a json file containing the dictionary `(r)qaoa.asdict()`.
Running the following code, a file named _{parent_uuid}--{uuid}--qaoa_test.json_ will be created (if parent_uuid is None, this will be omitted).
```
q = QAOA()
q.compile(problem)
q.optimize()
q.dump(file_name='qaoa_test', indent=1)
```

This method also accepts compression. Running the following code, the file created will be a `.gz `compressed file.
```
q.dump(file_name='qaoa_test', compresslevel=2)
```

###### keys not to include
 `.asdict`, `.dumps` and `.dump` methods accept the `'keys_not_to_inclue'` parameter. This parameter should be a list of keys (strings). The returned dictionaries (or jsons) will not include all those keys (and values) that has been passed through the `'keys_not_to_include'`.
As an example:
```
q.asdict(keys_not_to_inclue=['input_problem', 'input_parameters', 'results'])
```
will return a dictionary as the following:
```
{
   'header': {...},
   'data': {
      'exp_tags': {...}
   }
}
```
`'keys_not_to_inclue'` can also include keys of dictionaries that are inside other dictionaries, for example `'parent_uuid'`, which is inside `'identification'`. 

### Parameters to dict without keys starting with "_"
When `(r)qaoa.asdict()` method is called the input parameters objects (`CircuitProperties`, `BackendProperties`, `ClassicalOptimizer`, etc.) are converted to dictionaries using `dict(...)`, example:
````
{
...
'backend_properties': dict(self.backend_properties),
'classical_optimizer': dict(self.classical_optimizer),
...
}
````
To remove the "\_" from those keys that start with "\_", the following code has been introduced in the `Parameters` class:
```
class Parameters:
    def __iter__(self):
        for key, value in self.__dict__.items():
            # remove "_" from the beginning of the key if it exists
            yield (key[1:] if key.startswith("_") else key, value)
```
The same has been introduced in the problem class.

### Job id for simulators
The generation of a job id has been implemented for each simulator backend. This id is generated when the wavefunction (or counts) is computed. This way, we emulate the functionality of the QPUs.

Example:
```
class QAOAQiskitBackendShotBasedSimulator(QAOABaseBackendShotBased, QAOABaseBackendParametric):
...
    def get_counts(self, params: QAOAVariationalBaseParams) -> dict:
        ...
        # generate a job id for the wavefunction evaluation
        self.job_id = generate_uuid()
        ...
```

## TO DO
- [x] Impementation
> - [x] qaoa.results.asdict()
> - [x] rqaoa.results.asdict()
> - [x] (r)qaoa.asdict()
> - [x] (r)qaoa.dumps()
> - [x] (r)qaoa.dump()
> - [x] (r)qaoa.dump(compresslevel=x)
> - [x] (r)qaoa.set_exp_tags()
> - [x] (r)qaoa uuids
- [x] Tests
> - [x] qaoa.results.asdict()
> - [x] rqaoa.results.asdict()
> - [x] (r)qaoa.asdict()
> - [x] (r)qaoa.dumps()
> - [x] (r)qaoa.dump()
> - [x] (r)qaoa.dump(compresslevel=x)
> - [x] (r)qaoa.set_exp_tags()
> - [x] (r)qaoa uuids

## Questions
- When setting the `'parent_uuid'` up there should be a check with the database to check if this uuid is correct?

### CODE REVIEW

File: openqaoa/workflows/optimizer.py [hunk 1]
```diff
@@ -610,6 +610,17 @@ def set_rqaoa_parameters(self, **kwargs):
 
         return None
 
+    def set_exp_tag(self, **kwargs):
```
raulconchello, line 613: - [ ] A function like this one should be implemented for the QAOA object too.

raulconchello:
If you run Notebook 2 and try: `q_noisy_shot.dumps()` it doesn't work because there are some tuples as dictionary keys

raulconchello:
If you run Notebook 2 and try: `q_noisy_shot.dumps()` it doesn't work because there are some tuples as dictionary keys

openqaoa/rqaoa/rqaoa_results.py
```diff
@@ -12,119 +12,88 @@
...
-        Saves the result as json file.
-
-        Parameters
-        ----------
-        file_path : str
-            The name of the file to save the result. If None, the result is saved as 'result.json'.
-        indent : int
-            The number of spaces to indent the result in the json file. If None, the result is not indented.
+        Plots the correlation matrix of the i-th step of the RQAOA.
+        TODO : add more options
```
raulconchello, line 96: this method should be improved in another PR

raulconchello:
`rqaoa.results.plot_corr_matrix(step=..)` should be improved. I guess it would be better to do so in another PR.

openqaoa/optimizers/result.py
```diff
@@ -54,9 +54,11 @@ class Result:
     """
 
     def __init__(
-        self, log: Type[Logger], method: Type[str], cost_hamiltonian: Type[Hamiltonian]
+        self, log: Type[Logger], method: Type[str], cost_hamiltonian: Type[Hamiltonian], backend: str
     ):
 
+        self.__backend = backend
+
```
Q-lds, line 61: Why the extra space?

openqaoa/optimizers/result.py
```diff
@@ -54,9 +54,11 @@ class Result:
...
         self.method = method
```
Q-lds, line 63: Why the extra space?

openqaoa/optimizers/result.py
```diff
@@ -108,6 +110,48 @@ def __init__(
...
+
+        return_dict = {}
+        return_dict['method'] = self.method
+        if keep_cost_hamiltonian: return_dict['cost_hamiltonian'] = convert2serialize(self.cost_hamiltonian)
+        return_dict['evals'] = self.evals
+        return_dict['most_probable_states'] = self.most_probable_states
+
+        complx_to_str = lambda x: str(x) if isinstance(x, np.complex128) or isinstance(x, complex) else x
+        
+        if complex_to_string and issubclass(self.__backend, QAOABaseBackendStatevector):
```
Q-lds, line 138: The docstring should explain this conditional statement, as it is rather not trivial!

openqaoa/optimizers/result.py
```diff
@@ -108,6 +110,48 @@ def __init__(
...
+            return_dict['intermediate'] = {}
+            for key, value in self.intermediate.items():
+                return_dict['intermediate'][key] = [[complx_to_str(item) for item in list_] for list_ in value if (isinstance(list_, list) or isinstance(list_, np.ndarray))] \
```
Q-lds, line 141: Maybe format the whole expression to increase the readability?

openqaoa/optimizers/training_vqa.py
```diff
@@ -297,7 +297,7 @@ def results_dictionary(self,
         date_time = datetime.now().strftime("%d.%m.%Y_%H.%M.%S")
         file_name = f'opt_results_{date_time}' if file_name is None else file_name
         
-        self.qaoa_result = Result(self.log, self.method, self.vqa.cost_hamiltonian)
+        self.qaoa_result = Result(self.log, self.method, self.vqa.cost_hamiltonian, type(self.vqa))
```
Q-lds, line 300: So, result expects a `backend` variable which declared as a string. However, _in general With one argument, return the type of an object. The return value is a type object and generally the same object as returned by `object.__class__`_  (see  https://docs.python.org/3/library/functions.html#type)

Could this eventually be a problem? It may be worth double checking

openqaoa/workflows/optimizer.py
```diff
@@ -56,6 +59,18 @@ class Optimizer(ABC):
             A list containing the available cloud providers
         compiled: `Bool`
             A boolean flag to check whether the optimizer object has been correctly compiled at least once     
+        id: `dict`
+            A dictionary containing the identification stamps of the optimizer object. 
+            'uuid' is the unique identifier of the optimizer object, 
+            'parent_uuid' is the unique identifier of the parent object (e.g. the project to which the optimizer object belongs), the user can set this value using the set_identification method, 
+            'type' is the type of the object,
```
Q-lds, line 66: What's the type of the object?

openqaoa/workflows/optimizer.py
```diff
@@ -56,6 +59,18 @@ class Optimizer(ABC):
             A list containing the available cloud providers
         compiled: `Bool`
             A boolean flag to check whether the optimizer object has been correctly compiled at least once     
+        id: `dict`
+            A dictionary containing the identification stamps of the optimizer object. 
+            'uuid' is the unique identifier of the optimizer object, 
+            'parent_uuid' is the unique identifier of the parent object (e.g. the project to which the optimizer object belongs), the user can set this value using the set_identification method,
```
Q-lds, line 65: Maybe `experiment_uuid`? It will be more explicit that way?

openqaoa/workflows/optimizer.py
```diff
@@ -75,9 +90,47 @@ def __init__(self, device=DeviceLocal('vectorized')):
...
 
-    def asdict(self):
-        attributes_dict = convert2serialize(self)
-        return attributes_dict
+        # Initialize the identifier stamps, we initialize all the stamps needed to None
+        self.id = {
+            'uuid': None, #we generate a uuid for the experiment when we compile it
+            'parent_uuid': None, #user can set the parent uuid when they create the experiment using the set_identification method
+            'type': None, #type of the object
+            'datetime': datetime.now().strftime("%m/%d/%Y_%H:%M:%S")
```
Q-lds, line 98: Here we need to be careful. Let's make sure that ALL THE TIMES ARE IN UTC and the timezone is explicit recorded  by in the `strftime`. Also, better to change the order to the following `'%Y-%m-%dT%H:%M:%S.%f%z'`: the advantage is that they are all already time sorted (otherwise next to first of jan 2022 comes first of jan 2023 etc)

openqaoa/optimizers/result.py
raulconchello: done

openqaoa/optimizers/result.py
raulconchello: done

openqaoa/optimizers/training_vqa.py
raulconchello: shouldn't be declared as a string (changed). It shouldn't be a problem, but if someone changes  `QAOABaseBackendStatevector.__class__` should also change line 136, otherwise tests will fail

openqaoa/optimizers/result.py
raulconchello: done

openqaoa/optimizers/result.py
raulconchello: done

openqaoa/workflows/optimizer.py
raulconchello: RQAOA or QAOA

openqaoa/workflows/optimizer.py
raulconchello: I used `parent_uuid`, because if we are running a RQAOA the intermediate QAOA will have as a `parent_uuid` the uuid of the parent RQAOA.
But we can change the approach if you want.

openqaoa/workflows/optimizer.py
raulconchello: I have now used `'%Y-%m-%d_%H:%M:%S.%f%z'` and `datetime.utcnow()` this way the time is always in utc+0, but I am not sure that you meant that. 
Was it that, or to use the local time and save somehow the time zone?

openqaoa/workflows/optimizer.py
raulconchello: I have included it in that line

openqaoa/workflows/optimizer.py
Q-lds: No, that's fine :)

Q-lds:
`q.set_identification(parent_uuid='72d1f10f-8f38-4411-998d-76c2ceb95688')` shall we change this to `set_experiment_id(exp_id=....)`? This should make the usage of the field more clear

openqaoa/workflows/optimizer.py
Q-lds: We need both: we need a run_id, and an experiment_id.
- run_id is the id of the experiment. If qaoa, then run_id is job id of the qaoa you want to run; if rqaoa then the run_id is the overall rqaoa id, and each qaoa will have its own set of ids
- exp_id is the overall id for the experiment

I am happy to discuss finding better names for these two ids though!

tests/test_simulators.py
```diff
@@ -0,0 +1,59 @@
...
+
+class TestSimulators(unittest.TestCase):
+    
+    def test_job_ids(self):
+
+        #define problem
+        problem = QUBO.random_instance(3)
+
+        #loop over all simulators
+        for n in ['qiskit.qasm_simulator', 'qiskit.shot_simulator', 'qiskit.statevector_simulator', 'vectorized', 'pyquil.statevector_simulator']:
```
Q-lds, line 29: I think we should also have the tests for the QVM too

tests/test_utilities.py
```diff
@@ -797,6 +797,101 @@ def test_qaoa_probabilities(self):
...
+
+        # Expected dictionary
+        expected_dict = [
+            {'a': [1,2,3], 'b': {'x':1,'y':2}}, {'a': [1,2,3], 'b': '(1+1j)'}
+        ]
+
+        # Test that the dictionary has been generated correctly
+        assert serialized_dict == expected_dict, f'Object has not been converted correctly'
+
+    def test_generate_uuid(self):
```
Q-lds, line 873: We should also test:
- That all QAOA belonging to a RQAOA have the underlying RQAOA id
- Having assigned a project_id (or parent id) all children correctly feature it

openqaoa/workflows/optimizer.py
raulconchello: I used "parent_uuid" for both types of uuid because I was imagining a table in the database like this one:
https://www.notion.so/entropicawiki/OQ-database-e678ba6dd4d644eab40da2af00bcf0ed#b05158dcc5554bf5b5d2de85ef474728
With that you would be able to see the hierarchy.


However, depending on how we dump the data to the database all of this discussion is not useful. This is because, if we dump into the database at each RQAOA step then we need the id of the RQAOA in each QAOA. But, if we dump after the whole RQAOA, then we don't need the QAOA to contain any uuid because it will be in the RQAOA info.

tests/test_simulators.py
raulconchello: done

openqaoa/backends/simulators/qaoa_qiskit_sim.py
```diff
@@ -284,6 +288,9 @@ def qaoa_circuit(self, params: QAOAVariationalBaseParams) -> QuantumCircuit:
         qaoa_circuit: `QuantumCircuit`
             The final QAOA circuit after binding angles from variational parameters.
         """
+        # generate a job id for the wavefunction evaluation
+        self.job_id = generate_uuid()
```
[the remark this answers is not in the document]
raulconchello: done

tests/test_utilities.py
raulconchello: this we should test it, depending on how we decide to dump to the database, as I was saying here https://github.com/entropicalabs/openqaoa/pull/144#discussion_r1060344210 , no?

openqaoa/workflows/optimizer.py
Q-lds: The atomic experiment in openqaoa will be a QAOA. That is, all the measurement records for however many optimisations are needed to terminate the optimisation.

So, if RQAOA has 10 limitations we will save 10 files. Each QAOA will also need to reference the correct RQAOA id.

raulconchello: ok, but for now this is not implemented this way, because you can dump only when RQAOA is done, so we should rethink how we do the dumping, no?

Regarding the ids: would it make sense to have, "experiment_uuid" (the id of this workflow), "project_uuid" (the project id), and "parent_uuid" (this would be empty for RQAOA and single QAOA, but filled with "experiment_uuid" for QAOA steps in RQAOAs)

docs/source/conf.py
```diff
@@ -57,7 +57,7 @@
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = ['.docs/source/notebooks/community_tutorials/*']
+exclude_patterns = ['.docs/source/notebooks/community_tutorials/*', '.docs/source/notebooks/X_dumping_data.ipynb']
```
vishal-ph, line 60: This line stays or is removed depending on whether we keep/remove the notebook from the merge

openqaoa/workflows/optimizer.py
```diff
@@ -193,12 +273,155 @@ def set_classical_optimizer(self, **kwargs):
...
+
+        # we return a dictionary (serializable_dict) that will have two keys: header and data
+        serializable_dict = {
+            "header": header, # header is a dictionary containing all the data that can identify the experiment
+            "data": data, # data is a dictionary containing all the input and output data of the experiment (also the experiment tags)
+        }
+
+        return serializable_dict
+
+    def asdict(self, keys_not_to_include:List[str]=[]):
```
vishal-ph, line 339: recommendation: `exclude_keys`

openqaoa/workflows/optimizer.py
```diff
@@ -409,14 +637,43 @@ def optimize(self, verbose=False):
         if self.compiled == False:
             raise ValueError('Please compile the QAOA before optimizing it!')
 
+        # timestamp for the start of the optimization
+        self.header['execution_time_start'] = int(datetime.utcnow().timestamp())
```
vishal-ph, line 641: confirm whether this is always correct (i.e. some issues found on google (https://stackoverflow.com/questions/28573873/python-datetime-utcnow-returning-incorrect-datetime)

openqaoa/workflows/optimizer.py
```diff
@@ -741,21 +1010,26 @@ def optimize(self, verbose=False):
         full_solutions = rqaoa.final_solution(
             elimination_tracker, cl_ground_states, self.problem.hamiltonian)
 
+        # timestamp for the end of the optimization
+        self.header['execution_time_end'] = int(datetime.utcnow().timestamp())
```
vishal-ph, line 1014: also time issue

openqaoa/workflows/optimizer.py
raulconchello: it was correct because timestamp is defined as 

> that have elapsed since 00:00:00 [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time) on 1 January 1970

but I have change the library so there is no confusion

Q-lds (approved):

PR #147, opened 2022-12-07
Implementing the analytical formula for p=1 as a backened

kidiki:
### Summary of the changes
 - Implementing a new backend to obtain the expectation value of the quantum circuit analytically. 
 
 - Since this backend doesn't have the concept of measurement outcomes and the most probable string, self.measurement_outcomes passed as an empty dictionary. Also, the backend is given to the Result class so that we check if the backend is of type "QAOABackendAnalyticalSimulator" before trying to obtain the best string. 

- Another change regards the RQAOA and the function exp_val_hamiltonian_termwise in utilities.py which accepts 'analytical' as a param.  We set it to True if the backend is "QAOABackendAnalyticalSimulator". 

- Consider removing analytical = isinstance(qaoa_backend, QAOABaseBackendStatevector)

### Relevant context 
The analytical formula will allow for simulating much larger systems when p=1.

### Dependencies
No dependencies are required.

## How Has This Been Tested?

test/test_analytical_simulator.py


tests/test_analytical_simulator.py
```diff
@@ -0,0 +1,68 @@
...
+from openqaoa.utilities import random_k_regular_graph
+from openqaoa.workflows.optimizer import QAOA, RQAOA
+from openqaoa.problems.problem import QUBO, MaximumCut
+from openqaoa.devices import create_device
+
+"""
+A set of tests for FastQAOA: see the notebook Test_Examples.ipynb for explanations of how the
+expected answers are derived.
+"""
+def run_Rqaoa_experiment():
```
Q-lds, line 28: The tests are not correctly implemented: see any of the other test for the template (they need to be part of a class and they need to have names that begin with `test_`

tests/test_analytical_simulator.py
```diff
@@ -0,0 +1,68 @@
...
+    
+    g = random_k_regular_graph(degree=3, nodes=range(8), weighted=True, biases=False) 
+
+    # Define te problem and translate it into a binary Qubo.
+    maxcut_prob = MaximumCut(g)
+    maxcut_qubo = maxcut_prob.get_qubo_problem()
+
+    
+    # Define the RQAOA object (default rqaoa_type = 'adaptive')
+    R = RQAOA()
```
Q-lds, line 38: Always use lower case letters for variables!

kidiki: sure, will do, thanks

openqaoa/backends/simulators/qaoa_analytical_sim.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class QAOABackendAnalyticalSimulator(QAOABaseBackend):
+    r"""
+    TODO write something here.
+    """
+    
+    def __init__(self,
+                 circuit_params: QAOACircuitParams,
+                 prepend_state: Optional[Union[np.ndarray, List[complex]]],
```
vishal-ph, line 38: arguments in lines 38-41 are not needed

tests/test_analytical_simulator.py
kidiki: I have implemented the changes you suggested. Can you take a look here: https://github.com/entropicalabs/openqaoa/blob/5a51f2c8c4f2fc5deaf68fd7583ee9afefe3f7fd/tests/test_analytical_simulator.py#L53

openqaoa/backends/simulators/qaoa_analytical_sim.py
```diff
@@ -0,0 +1,110 @@
...
+                 append_state: Optional[Union[np.ndarray, List[complex]]],
+                 init_hadamard: bool,
+                 cvar_alpha: float = 1):
+        
+        assert cvar_alpha == 1,  "Please use the shot-based simulator for simulations with cvar_alpha < 1"
+        
+        QAOABaseBackendStatevector.__init__(self, circuit_params,
```
vishal-ph, line 45: `QAOABaseBackendStatevector` -> `QAOABaseBackend`

kidiki:
My first PR! I only had to ask half of the company for help lol

openqaoa/backends/simulators/qaoa_analytical_sim.py
```diff
@@ -0,0 +1,111 @@
...
+        params: `QAOAVariationalBaseParams`
+            The QAOA parameters - an object of one of the parameter classes, containing 
+            variable parameters.
+
+        Returns
+        -------
+        float:
+            Expectation value of cost operator wrt to the QAOA parameters according to the analytical expression for p=1.
+        """
+        betas = params.betas
```
Q-lds, line 90: This is going to work only for the standard parametrisation. Is this wanted?

openqaoa/optimizers/result.py
```diff
@@ -54,9 +54,10 @@ class Result:
     """
 
     def __init__(
-        self, log: Type[Logger], method: Type[str], cost_hamiltonian: Type[Hamiltonian]
+        self, log: Type[Logger], method: Type[str], cost_hamiltonian: Type[Hamiltonian], backend: str
```
Q-lds, line 57: You also need to change the docs of the class. But did you agree with @raulconchello on this? Is this change consistent with the overall refactor of the result object?

openqaoa/optimizers/result.py
```diff
@@ -89,11 +90,10 @@ def __init__(
...
                 if len(log.job_ids.best) != 0 else []
         }
-
+        
         self.most_probable_states = most_probable_bitstring(
-            cost_hamiltonian, self.get_counts(log.measurement_outcomes.best[0])
-        ) if log.measurement_outcomes.best != [] else []
-
+                cost_hamiltonian, self.get_counts(log.measurement_outcomes.best[0])
+            ) if backend.__name__ != "QAOABackendAnalyticalSimulator" and log.measurement_outcomes.best != [] else []
```
Q-lds, line 96: If backend is already a string, why are you accessing `backend.__name__` ?

openqaoa/optimizers/training_vqa.py
```diff
@@ -297,7 +297,7 @@ def results_dictionary(self,
         date_time = datetime.now().strftime("%d.%m.%Y_%H.%M.%S")
         file_name = f'opt_results_{date_time}' if file_name is None else file_name
         
-        self.qaoa_result = Result(self.log, self.method, self.vqa.cost_hamiltonian)
+        self.qaoa_result = Result(self.log, self.method, self.vqa.cost_hamiltonian, type(self.vqa))
```
Q-lds, line 300: Maybe here you should try the `__name__` attribute? It is a bit confusing (for others) that the type of a VQA is mapped to a variable called backend. What do you think about it?

openqaoa/optimizers/result.py
raulconchello: Yes, we did that together

openqaoa/optimizers/result.py
raulconchello: Instead of comparing the name we could do `isintance(backend, QAOABackendAnalyticalSimulator)`

openqaoa/optimizers/result.py
Q-lds: What is it in your PR?

openqaoa/backends/simulators/qaoa_analytical_sim.py
```diff
@@ -0,0 +1,111 @@
...
+     
+     Parameters
+    ----------
+    circuit_params: QAOACircuitParams
+        An object of the class ``QAOACircuitParams`` which contains information on 
+        circuit construction and depth of the circuit. Note that it only works for p=1 and the X Mixer Hamiltonian.
+    """
+    
+    def __init__(self, circuit_params: QAOACircuitParams, 
+                 prepend_state = None,
```
raulconchello, line 43: These parameters are not used, I think `def __init__(self, circuit_params: QAOACircuitParam):` should be enough, otherwise, it can be confusing.

openqaoa/optimizers/result.py
Q-lds: It depends on the datatype of the backend. I think passing a string may be leaner than passing the object, but either works

openqaoa/optimizers/training_vqa.py
raulconchello: For the asdict method of PR #144 I need it to be `type(self.vqa)`, to check its parent, so`__name__` would work.

openqaoa/backends/simulators/qaoa_analytical_sim.py
vishal-ph: So, each parameterization has `betas` and `gammas` attributes. Since the expectation value formula only depends on `betas`, `gammas`, one can extract these values from any parameterization to plug into the formula and obtain the cost function.

vishal-ph: @kidiki, one thing to ensure is what happens in the case of `Extended` (including `FourierExtended`) and `WithBias` parameterizations, which may not have exactly these attributes but rather `gammas_singles`, `gammas_pairs`, `betas_singles`, `betas_pairs`

kidiki: Is it enough to assert that the params must be standard for now? and then later we can re-derived the analytical formula for extended?

vishal-ph: AFAIK, the generalization to Extended shouldn't be too hard. Let's make an attempt (perhaps @ezerc can help here), and if that's not the case then we restrict it to support only Standard.

openqaoa/backends/simulators/qaoa_analytical_sim.py
```diff
@@ -0,0 +1,119 @@
...
+import numpy as np
+
+from ...basebackend import QAOABaseBackend, QuantumCircuitBase
+from ...qaoa_parameters.baseparams import QAOAVariationalBaseParams
+from ...qaoa_parameters.operators import Hamiltonian
+from openqaoa.utilities import energy_expectation_analytical, generate_uuid
+
+# not sure about those ones:
+from ...qaoa_parameters.baseparams import QAOACircuitParams, QAOAVariationalBaseParams 
+from typing import Union, List, Tuple, Type, Optional
```
Q-lds, line 27: From PEP8 (see https://peps.python.org/pep-0008/#imports)

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.
> Imports should be grouped in the following order:
>  - Standard library imports.
>  - Related third party imports.
>  - Local application/library specific imports.

Let's start by making sure this folder follows the right convention. (Worth checking it black does it automatically)

openqaoa/backends/simulators/qaoa_analytical_sim.py
```diff
@@ -0,0 +1,119 @@
+#   Copyright 2022 Entropica Labs
```
Q-lds, line 1: Since this is a new file, maybe format it right away with https://github.com/psf/black ?

openqaoa/backends/simulators/qaoa_analytical_sim.py
```diff
@@ -0,0 +1,119 @@
...
+    
+    def expectation_w_uncertainty(self, params):
+        raise NotImplementedError("Not implemented yet. In progress.")
+        
+    def reset_circuit(self):
+        raise NotImplementedError("This method is irrelevant for this backend")
+    
+    def circuit_to_qasm(self):
+        raise NotImplementedError("This method is irrelevant for this backend")
+
```
Q-lds, line 112: Keep only a single newline! :)

tests/test_analytical_simulator.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        Testing if the analytical backend fails if the number of layers, p, is different than 1. 
+        """
+        exception = False
+        try:
+            n_qubits = 8
+            p = 2
+            mixer_hamil = X_mixer_hamiltonian(n_qubits)
+            register, cost_hamil, qaoa_circuit_params, variate_params = Disagrees_SetUp(n_qubits, p, mixer_hamil)
+            backend_analytical = QAOABackendAnalyticalSimulator(qaoa_circuit_params)
```
Q-lds, line 79: Codecov says that this line has never been tested: this seems to imply that the test failed BEFORE reaching line 79. Is this the desired behaviour?

tests/test_analytical_simulator.py
```diff
@@ -0,0 +1,186 @@
...
+    def test_different_mixer_fails(self):
+        """
+        Testing if the analytical backend fails if the mixer hamiltonian is different than X. 
+        """
+        exception = False
+        try:
+            n_qubits = 8
+            p = 1
+            mixer_hamil = XY_mixer_hamiltonian(n_qubits)
+            register, cost_hamil, qaoa_circuit_params, variate_params = Disagrees_SetUp(n_qubits, p, mixer_hamil)
```
Q-lds, line 94: Again, lines 94-95 were not executed by the test. Is this correct?

tests/test_analytical_simulator.py
kidiki: Ah, good catch! It was failing before that indeed. should be fixed now!

openqaoa/backends/simulators/qaoa_analytical_sim.py
kidiki: Formatted it with black, thanks!

tests/test_analytical_simulator.py
kidiki: It has been fixed, thanks!

vishal-ph (approved):

PR #152, opened 2022-12-15
Metadata for Problem objects

raulconchello:
## Description

Implementation of the attributes `qubo.metadata` and `qubo.problem_instance` in the QUBO class.

### Metadata
`qubo.set_metadata(metadata={...})`
This method adds the items of the input dictionary to `qubo.metadata` (which is a dictionary).
By default the metadata dictionary is empty, the users can fill it with any data that they want (data needs to be serializable).

##### Problem class
The property `problem_intance` has been included in the `Problem` class:
```
class Problem(ABC):
...
    @property
    def problem_instance(self):
        ...
        return {**{"problem_type": self.__name__}, **dict(self)}
```
this property is a dictionary of the serialization of the object + the problem type name (e.g. for number partitoning `problem_type='number_partitoning'`)

To get a good serialization of the object when calling `dict(self)`, the following has been included:
```
class Problem(ABC):
...
    def __iter__(self):
        for key, value in self.__dict__.items():
            # remove "_" from the beginning of the key if it exists
            new_key = key[1:] if key.startswith("_") else key
            # convert networkx graphs to dictionaries
            new_value = nx.node_link_data(value) if isinstance(value, nx.Graph) else value

            yield (new_key, new_value)
```

And to get the correct name when calling `self.__name__`, the `__name__` has been included in each problem class, example:
```
class NumberPartition(Problem):
...
    __name__ = "number_partition"
   ...
```

Then when the `QUBO` is created from the problem object the `self.problem_instance` is passed, example:
```
class NumberPartition(Problem):
...
    def get_qubo_problem(self):
        ...
        return QUBO(self.n_numbers, terms, weights, self.problem_instance)
```

##### QUBO class

The initialization of the `QUBO` class has now a new parameter `problem_instance`, and an attribute is created with the value:
```
class QUBO:
...
    def __init__(self, n, terms, weights, 
                    problem_instance:dict={"problem_type": "generic_qubo"}, 
                    ...):
        ...
        self.problem_instance = problem_instance
```
therefore `qubo.problem_instance` will be `{"problem_type": "generic_qubo"}` if nothing is passed, but when wubo is created from one `problem` this dictionary will have the problem type name and the serialization of the problem.

##### Examples

Number Partitoning
```
NumberPartition_problem = NumberPartition.random_instance(n_numbers=3)
NumberPartition_qubo = NumberPartition_problem.get_qubo_problem()

dict(NumberPartition_qubo)
```

> _out:_
 ```
{'terms': [[0, 1], [0, 2], [1, 2]],
 'weights': [72.0, 12.0, 12.0],
 'constant': 73,
 'n': 3,
 'problem_instance': {
           'problem_type': 'number_partition',
           'numbers': [6, 6, 1],
           'n_numbers': 3},
 'metadata': {}}
```

Minimum Vertex Cover
```
MinimumVertexCover_problem = MinimumVertexCover.random_instance(n_nodes=3, edge_probability=0.5)
MinimumVertexCover_qubo = MinimumVertexCover_problem.get_qubo_problem()
MinimumVertexCover_qubo.set_metadata({'tag1': 'value1', 'tag2': 'value2'})

dict(MinimumVertexCover_qubo)
```
> _out_:
```
{'terms': [[0, 2], [1, 2], [0], [1], [2]],
 'weights': [2.5, 2.5, 2.0, 2.0, 4.5],
 'constant': 6.5,
 'n': 3,
 'problem_instance': {
          'problem_type': 'minimum_vertex_cover',
          'G': {
                    'directed': False,
                    'multigraph': False,
                    'graph': {},
                    'nodes': [{'id': 0}, {'id': 1}, {'id': 2}],
                    'links': [{'source': 0, 'target': 2}, {'source': 1, 'target': 2}]},
          'field': 1.0,
          'penalty': 10},
 'metadata': {'tag1': 'value1', 'tag2': 'value2'}}
```

QUBO
```
qubo = QUBO.random_instance(3)

dict(qubo)
```
>_out:_
```
{'terms': [[2], [0, 1], [0, 2]],
 'weights': [-0.833099253577807, 14.405500837522837, -1.0046728497563495],
 'constant': 0,
 'n': 3,
 'problem_instance': {'problem_type': 'generic_qubo'},
 'metadata': {}}
```

## TO DO
- [x] Have default keys and values?
- [x] Have required keys?
- [x] Write test

### CODE REVIEW

File: openqaoa/problems/problem.py [hunk 1]
```diff
@@ -118,6 +118,10 @@ def __init__(self, n, terms, weights, clean_terms_and_weights=False):
         self.constant = constant
         self.n = n
 
+        self.name = "generic_qubo" if "name" not in metadata else metadata["name"]
```
raulconchello, line 121: Is this attribute needed? this info is already in the metadata.

Or should be removed from the metadata?

openqaoa/problems/problem.py
```diff
@@ -583,6 +602,9 @@ def __init__(self, numbers=None):
         self.numbers = numbers
         self.n_numbers = None if numbers == None else len(self.numbers)
```
Q-lds, line 604: Let's try to remove all the inconsistent spacing too

openqaoa/problems/problem.py
raulconchello: But, do we keep both `my_problem.name` and `my_problem.metadata.name`?

openqaoa/problems/problem.py
```diff
@@ -135,6 +139,17 @@ def n(self, input_n):
...
+    def set_metadata(self, metadata:dict={}):
+        """
+        Sets the metadata of the problem.
+
+        Parameters
+        ----------
+        metadata: dict
+            The metadata of the problem. All keys and values will be stored in the metadata dictionary.
+        """
+        self.metadata = {**self.metadata, **metadata}
```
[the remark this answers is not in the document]
raulconchello: How can we check that? By a `try` and `except` and trying to create a json?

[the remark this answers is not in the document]
Q-lds: I'd probably just check that the datatype of the dictionary's key is always a string, and that the value is always a python primitive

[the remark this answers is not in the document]
raulconchello: Done

openqaoa/problems/problem.py
raulconchello: I am not sure which spacing you are referring to here

openqaoa/problems/problem.py
```diff
@@ -321,6 +336,10 @@ def __init__(self,
         self.A = A if A else 2 * distance_matrix.max()
         self.B = B
 
+        # set the metadata
+        self.metadata = {'name': 'tsp', 'n_cities': self.n_cities}
```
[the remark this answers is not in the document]
raulconchello: do you mean why `n_cities` is by default in the metadata? 
Because, when the QUBO is created this metadata is passed as a parameter.

However, I am not sure that `n_cities` here makes sense because `n` will be equal to `n_cities` (and the same happens for all the other type of problems, with `n_items`, `n_nodes`, `n_numbers`, etc.). Should we remove these attributes from the metadata?

openqaoa/problems/problem.py
Q-lds: The problem is that we are being inconsistent: why are we storing only some of the parameters and not others? And why are they being duplicated?

There are two type of data we care about when it comes to problems:
- the internal ones are all already stored as attributes of the object
- the external are the metadata tags that the users adds (e.g. the type of graph, the weighting strategy, etc)

I think that at the level of the object, we don't need to add the attributes to the metadata.

However, we need to make sure that when we get the input problem in the result of a workflow. There we need all the attributes (serialised) and the metadata. Whether we want things here to have a different structure (e.g. `n_cities` is stored in the metadata) can be discussed

openqaoa/problems/problem.py
```diff
@@ -135,6 +140,25 @@ def n(self, input_n):
...
+        ----------
+        metadata: dict
+            The metadata of the problem. All keys and values will be stored in the metadata dictionary.
+        """
+        # check if the keys are strings and the values are python primitives (so that they can be serialized)
+        for key, value in metadata.items():
+            if not isinstance(key, str):
+                raise TypeError("The keys of the metadata dictionary must be strings. {} is not a string.".format(key))
+            if not isinstance(value, (str, int, float, bool)):
+                raise TypeError("The values of the metadata dictionary must be python primitives. {} is not a python primitive.".format(value))
```
Q-lds, line 157: Let's double check what are the primitives!

openqaoa/problems/problem.py
```diff
@@ -135,8 +168,26 @@ def n(self, input_n):
...
+        Parameters
+        ----------
+        metadata: dict
+            The metadata of the problem. All keys and values will be stored in the metadata dictionary.
+        """
+        #here we try if the dictionary is serializable
+        try:
+            _ = json.dumps(metadata)
+        except Exception as e:
+            raise e
```
raulconchello, line 184: I think it is easier to check if the metadata is serializable by doing this, @Q-lds what do you think?

openqaoa/problems/problem.py
```diff
@@ -42,6 +43,25 @@ def random_instance(**kwargs):
...
+    def __iter__(self):
+        for key, value in self.__dict__.items():
+            # remove "_" from the beginning of the key if it exists
+            new_key = key[1:] if key.startswith("_") else key
+            # convert networkx graphs to dictionaries
+            new_value = nx.node_link_data(value) if isinstance(value, nx.Graph) else value
+            yield (new_key, new_value)
+
+    @property
+    def problem_instance(self):
```
raulconchello, line 55: perhaps it is not the best name?

openqaoa/problems/problem.py
```diff
@@ -118,6 +140,17 @@ def __init__(self, n, terms, weights, clean_terms_and_weights=False):
         self.constant = constant
         self.n = n
 
+        # attribute to store the problem instance
+        self.problem_instance = problem_instance
```
raulconchello, line 144: again, is the name correct?

openqaoa/problems/problem.py
```diff
@@ -118,6 +140,28 @@ def __init__(self, n, terms, weights, clean_terms_and_weights=False):
...
+            # remove "_" from the beginning of the key if it exists
+            yield (key[1:] if key.startswith("_") else key, value)
+
+    def __setattr__(self, __name, __value):
+        # check if problem_instance is json serializable, also check if metadata is json serializable
+        if __name == "problem_instance" or __name == "metadata":
+            try:
+                _ = json.dumps(__value)
+            except Exception as e:
+                raise e
```
raulconchello, line 160: I think it is easier to check if the metadata is serializable by doing that, @Q-lds what do you think?

Q-lds: Sure, good idea!

tests/test_problems.py
```diff
@@ -796,6 +839,38 @@ def test_assertion_fn():
...
+    def test_problem_instance(self):
+        """
+        Test problem instance method of the QUBO class.
+        We create a random instance of all the different problems, we generate the QUBO problem out of it and then we check if the problem instance attribute is correct, by comparing the keys of the problem instance with the expected keys.
+        """
+
+        all = {
+            "tsp":TSP.random_instance(n_cities=3),
+            "number_partition":NumberPartition.random_instance(n_numbers=3),
+            "maximum_cut":MaximumCut.random_instance(n_nodes=3, edge_probability=0.5),
```
Q-lds, line 851: can you put all both `n` and the `edge_prob` to take random values? It may be useful in the future, as they may give some edge cases :)

raulconchello: done

Q-lds:
Since `problem_instance` will have varying length depending on the underlying problem, we need to change its data structure so that it is machine parsable.

So, if some of these fields are always present then then can be keys-values pairs in a dictionary. Otherwise, we need to turn them into lists. For example, `"nodes": [{"id": 0}, {"id": 1}, {"id": 2}]` is generated this way because you don't know a priory how many nodes we will have.

openqaoa/problems/problem.py
vishal-ph: would you rather like to call it dumps?

Q-lds (approved):

vishal-ph (approved):

PR #153, opened 2022-12-15
Refactoring of the Input Parameters: Classical Optimizer, Backend, etc.

raulconchello:
## TO DO
- [x] $\alpha$. Consistent documentation between Object classes of Input Params and workflows
- [x] $\beta$. Remove the `_p` names (removed in #144)
- [ ] $\gamma$. Fix spelling and finalise the naming convention
- [ ] $\delta$. We should not have things like `annealing_time` if we are using standard parametrisation
- [ ] $\epsilon$. Input consistency checks needs to be updated (this may be complex, and we need to find a good solution)
- [ ] $\zeta$. Make use of the Enum feature in order to deal with `string` variables which belong to a finite set (eg, `param_type` can only take fixed values)


raulconchello:
_We should not have things like annealing_time if we are using standard parametrization._

I see two solutions for that: 
- [ ] we only create the attributes that are not `None`. The code below is an example of how we can do it. This code is taking all the parameters of the initialization (`optimize`, `method`, `maxiter`, ...) and create an attribute with this name only if the value is not `None`.
```
import inspect

class ClassicalOptimizer(Parameters):
...
    def __init__(self,
                 optimize: bool = True,
                 method: str = 'cobyla',
                 maxiter: int = 100,
                 maxfev : int = None,
                 ....):
       for param_name in inspect.signature(ClassicalOptimizer).parameters:
            if eval(param_name) is not None:
                setattr(self, param_name, eval(param_name))
```
or
- [ ] when we return the dictionary of the object the attributes that are none.

raulconchello:
## $\epsilon$. _Input consistency checks need to be updated (this may be complex, and we need to find a good solution)_

```
    @property
    def param_type(self):
        return self._param_type

    @param_type.setter
    def param_type(self, value):
        if value not in ALLOWED_PARAM_TYPES:
            raise ValueError(
                f"param_type {value} is not recognised. Please use {ALLOWED_PARAM_TYPES}")
        self._param_type = value
```

raulconchello:
## $\zeta$. _Make use of the Enum feature in order to deal with `string` variables which belong to a finite set (eg, `param_type` can only take fixed values)_

openqaoa/workflows/parameters/qaoa_parameters.py
```diff
@@ -17,19 +17,19 @@
 import qiskit
 from scipy.optimize._minimize import MINIMIZE_METHODS
 import scipy
-import numpy
+# import inspect
```
vishal-ph, line 20: commented out import. Not needed?

PR #154, opened 2022-12-15
Eliminate white spaces in dict keys in Results

raulconchello:
Refactoring of the QAOA Results class. For all the result dictionaries keys with white spaces were used. Those are changed to "_" here.

This PR can brake old code

## TO DO
- [ ] Check that it has been changed everywhere, where these keys were used
- [ ] Write some test?


openqaoa/optimizers/result.py
```diff
@@ -62,30 +62,30 @@ def __init__(
...
+            "number_of_evals": log.func_evals.best[0],
+            "jac_evals": log.jac_func_evals.best[0],
+            "qfim_evals": log.qfim_func_evals.best[0],
         }
 
         self.intermediate = {
-            'angles log': np.array(log.param_log.history).tolist(),
-            'intermediate cost': log.cost.history,
-            'intermediate measurement outcomes':
+            'angles_log': np.array(log.param_log.history).tolist(),
```
Q-lds, line 71: Shall we rename this to `intermediate_angles`? @vishal-ph

vishal-ph: @Q-lds , @raulconchello. So one point about naming these fields. Since the outer field is already called intermediate, maybe we should rename all of them to just "cost", "measurement_outcomes", "angles" (getting rid of the word intermediate)

Similarly for the field optimized, lets remove the optimized from the sub fields as well. It's redundant in my opinion.

raulconchello: It is redundant, but maybe they can be useful at some point. For example, when you are looking at a json of the results, the file is very long, so it can be helpful if the keys are "intermediate_..." or "optimized_...", so you know what you are looking at without the need of scrolling up to the parent. 
I think it doesn't hurt.

Q-lds: These fields are supposed to be accessed programmatically so I am in favor of what Vishal is saying. 

I also have another question: how is the optimized set chosen? The lowest value in the optimization? Or the last? If it is the last, then the true cost-history will be `intermediate.cost` + `optimized.cost` right? 

I think we need to be a bit more explicit with this: how about adding what optimisation round does the optimized result corresponds to?

openqaoa/optimizers/result.py
```diff
@@ -62,30 +62,30 @@ def __init__(
...
-            'optimized angles':
+            'optimized_angles':
                 np.array(log.param_log.best[0]).tolist()
                 if log.param_log.best != [] else [],
-            'optimized cost':
+            'optimized_cost':
                 log.cost.best[0]
                 if log.cost.best != [] else None,
-            'optimized measurement outcomes':
+            'optimized_measurement_outcomes':
```
Q-lds, line 85: As a general remark: I still think that having an `optimized_measurement_outcomes` and getting a wave-function is quite misleading. But I think we can probably still live with it ...

openqaoa/optimizers/result.py
```diff
@@ -62,30 +62,30 @@ def __init__(
...
                 log.measurement_outcomes.best[0]
                 if log.measurement_outcomes.best != [] else {},
-            'optimized run job id': 
+            'optimized_run_job_id':
```
Q-lds, line 88: Again, if we are (ever) to select the lowest cost function (and not the last) then this field is correlated with the iteration counter

openqaoa/optimizers/training_vqa.py
```diff
@@ -283,15 +283,15 @@ def results_dictionary(self,
                 #. "solution"
                     #. "bitstring"
                     #. "degeneracy"
-                #. "number of evals"
-                #. "jac evals"
-                #. "qfim evals"
+                #. "number_of_evals"
+                #. "jac_evals"
+                #. "qfim_evals"
                 #. "parameter log"
```
Q-lds, line 289: Why is there a space here?

raulconchello: I think these shouldn't be there, they are not in the results

raulconchello: I eliminated that part because in that function we are not returning anything

openqaoa/optimizers/result.py
raulconchello: I also think it is a little misleading, but how could we change that? I don't think it is a good idea to change the key depending on the type of backend

openqaoa/optimizers/result.py
raulconchello: @shahidee44 will respond better than I can. 
Can we include the optimization round?

openqaoa/optimizers/result.py
raulconchello: Yes, if we can include the round it will refer to this value too

openqaoa/optimizers/result.py
raulconchello: > @Q-lds , @raulconchello. So one point about naming these fields. Since the outer field is already called intermediate, maybe we should rename all of them to just "cost", "measurement_outcomes", "angles_log" (getting rid of the word intermediate)
> 
> Similarly for the field optimized, lets remove the optimized from the sub fields as well. It's redundant in my opinion.

Ok, I have updated the keys to be the same for both dictionaries `intermediate` and `optimized`:
```
        self.intermediate = {
            'angles': ...,
            'cost': ...,
            'measurement_outcomes': ...,
            'job_id': ...,
        }

        self.optimized = {
            'angles': ...,
            'cost': ...,
            'measurement_outcomes': ...,
            'job_id': ...,
        }
```

raulconchello: > These fields are supposed to be accessed programmatically so I am in favor of what Vishal is saying.
> 
> I also have another question: how is the optimized set chosen? The lowest value in the optimization? Or the last? If it is the last, then the true cost-history will be `intermediate.cost` + `optimized.cost` right?
> 
> I think we need to be a bit more explicit with this: how about adding what optimisation round does the optimized result corresponds to?

@shahidee44 will respond better than I can. 
Can we include the optimization round?

openqaoa/optimizers/result.py
Q-lds: No, it isn't. Let's leave it like this for the moment

openqaoa/optimizers/result.py
Q-lds: In the optimized result, yes. I think so!

shahidee44: Yes we can. Although, i'd be very careful about what we call that value. We can always repurpose the evaluations logger variable for such a task. Or just create a new variable for it if its not possible.

raulconchello: I have done the following to have `q.results.optimized['eval_number']` (an int which tells the evaluation in which we have found the optimized values)
https://github.com/entropicalabs/openqaoa/blob/a8e14d9d0f3e9d2dcfe555a682817c5fcbad0e01/openqaoa/optimizers/training_vqa.py#L230

Q-lds: @vishal-ph ?

openqaoa/optimizers/training_vqa.py
```diff
@@ -212,13 +218,17 @@ def optimize_this(self, x):
...
         log_dict.update({'cost': callback_cost})
+
         current_eval = self.log.func_evals.best[0]
         current_eval += 1
         log_dict.update({'func_evals': current_eval})
-        
+
+        eval_number = self.log.eval_number.history[-1] if len(self.log.eval_number.history) > 0 else 0
+        eval_number += 1
+        log_dict.update({'eval_number': eval_number})  #this one will say which evaluation is the optimized one
```
Q-lds, line 230: @vishal-ph in light of what we were discussing yesterday, could you check this?

openqaoa/optimizers/result.py
vishal-ph: Now that the functionality is already added, lets keep it.

vishal-ph: @raulconchello, you can resolve this thread.

openqaoa/optimizers/training_vqa.py
shahidee44: If the goal here is to just record the evluation number where the cost is the lowest.
Then all we need to do here is
```
log_dict.update({'eval_number': current_eval})
```

`history_update_bool` for `eval_number` can be `False`

If the cost is not updated, `eval_number` does not get updated, even if a value is passed in,.

tests/test_results.py
```diff
@@ -98,6 +100,20 @@ def _test_results_dict(self, results_dict):
         for key in expected_keys:
             assert key in results_dict.keys(), '{} key not in results dictionary'.format(key)
 
+    #test eval_number 
+    def test_qaoa_result_eval_number(self):
```
Q-lds, line 104: As a general rule, a viable test for this feature would be:
- for a deterministic optimizer check that the returned `results.optimized['eval_number']` is the same as the eval corresponding to the lowest value in the cost history.

Can we implement it?

Q-lds (approved):

openqaoa/optimizers/training_vqa.py
```diff
@@ -212,13 +218,15 @@ def optimize_this(self, x):
...
-        
+        callback_cost = self.vqa.expectation(self.variational_params)        
         log_dict.update({'cost': callback_cost})
+
         current_eval = self.log.func_evals.best[0]
         current_eval += 1
         log_dict.update({'func_evals': current_eval})
-        
+
+        log_dict.update({'eval_number': (current_eval - self.log.jac_func_evals.best[0] - self.log.qfim_func_evals.best[0])})  #this one will say which evaluation is the optimized one
```
vishal-ph, line 228: do we also need to subtract `self.log.hess_func_evals` or something similar? (For hessian evaluations, if specified)

Q-lds: Maybe we should have more tests? Indeed I think Vishal is right, and the fact that the tests did not pick it up it means that the test can be improved

raulconchello: I don't think `self.log.hess_func_evals` exists. When a hessian is called we update `self.log.jac_func_evals` too.

raulconchello: I think the value is fine, but I will improve the tests, using different optimization methods.

vishal-ph: Great, thanks for the clarification!

PR #158, opened 2022-12-27
Bugfix vanishing instances after elimination

kidiki:
## Description
Summary: Solving and RQAOA bug when vanishing problems. 
Context: unweighted graphs have a lot of edge cases when reduced during the RQAOA algorithm. 
Bug description: 
Approach to fixing: check if the new problem is an empty dictionary, which would mean that a total elimination happened. If so, set the new problem to the old one and break the optimization cycle. Hence, solving for the smallest non-vanishing instance or equivalently, increasing the cutoff. 

Note: this might cause a potential problem if the smallest non-vanishing instance is too big to be solved classically. TODO

## Checklist

## Type of change

## How Has This Been Tested?
- [ ] Write unit tests for the specific instances which were failing previously.


kidiki:
"Spiders are the only web developers that enjoy finding bugs." + Kristina ;) :D

kidiki:
Note that this bug occurs for 27 out of 1100 random instances for unweighted graphs with p=0.7 probability for an edge.

openqaoa/workflows/optimizer.py
```diff
@@ -713,26 +713,29 @@ def optimize(self, verbose=False):
             spin_map = rqaoa.spin_mapping(problem, max_terms_and_stats)
             # Eliminate spins and redefine problem
             new_problem, spin_map = rqaoa.redefine_problem(problem, spin_map)
+            if new_problem == problem:
+                print("Eliminations lead to a total reduction of the problem.\n Increasing the cutoff to solve the smallest non-vanishing problem.")
+                break
+            else:
```
raulconchello, line 719: this `else` is not needed, and I would say that without it the code is more readable, since `new_problem == problem` is a rare case

kidiki: Good point! Thanks! Done!

openqaoa/workflows/optimizer.py
```diff
@@ -727,7 +731,7 @@ def optimize(self, verbose=False):
 
             # problem is updated
             problem = new_problem
-            
+
```
Q-lds, line 734: Remove the new line

openqaoa/workflows/optimizer.py
```diff
@@ -713,7 +713,11 @@ def optimize(self, verbose=False):
             spin_map = rqaoa.spin_mapping(problem, max_terms_and_stats)
             # Eliminate spins and redefine problem
             new_problem, spin_map = rqaoa.redefine_problem(problem, spin_map)
-
+            # In case eliminations cancel out the whole graph, break the loop (solve classically) before reaching the predefined cutoff.
+            if new_problem == problem:
```
Q-lds, line 717: Is this always going to be correct? I guess the constant term may change, or what happens if symmetric relabeling happens?

openqaoa/workflows/optimizer.py
kidiki: that's not my code but sure.

openqaoa/workflows/optimizer.py
kidiki: It comes from the rqaoa.py file: I'm passing the old problem only if the new problem dict is empty.What do you mean by symmetric relabeling?

https://github.com/entropicalabs/openqaoa/blob/851352621182aa73ef998f722f5da9fa7551b486/openqaoa/rqaoa/rqaoa.py#L502

Q-lds: Assume the old problem has [0,1] and corresponding weight [1], and the new problem has [1,0] and [1]. Wouldn't this be the same graph? But the equality would fail.

Don't focus on my example: I was just wondering if the code would break in cases where you have, for example equivalent graphs but different objects

kidiki: I'd say no because the way the code is written the problem is always defined in increasing order of the nodes. But will discuss with Eze and Vishal after the holidays, very likely there are edge cases I haven't considered.

vishal-ph: @Q-lds Yes, @kidiki, you are right. This was an issue Eze and I faced before and decided to represent the nodes in increasing order.

openqaoa/rqaoa/rqaoa.py
```diff
@@ -581,10 +612,74 @@ def final_solution(elimination_tracker: list, cl_states: list, hamiltonian: Hami
...
+    spin_map: `dict`
+        Spin map containing the correlations and eliminations of the smallest non vanishing problem statement.
+    hamiltonian: `Hamiltonian`
+        Hamiltonian object containing the smallest non vanishing problem statement.
+
+    Returns
+    -------
+    cl_energy: `float`
+        The energy of the first solution wrt the cost Hamiltonian.
+
```
Q-lds, line 643: Remove the new line

vishal-ph: let me see if I can take care of this

vishal-ph: done

vishal-ph (approved):

PR #160, opened 2022-12-29
Removed unused imports and Added annotations

shahidee44:
## Description

- Removed unused imports and Added `from __future__ import annotations` 

## Checklist

## Type of change

## How Has This Been Tested?

test_all.py was used locally.

### DIFF
--- a/openqaoa/backends/qaoa_backend.py
+++ b/openqaoa/backends/qaoa_backend.py
@@ -11,6 +11,8 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+from __future__ import annotations
+
 from typing import Union, Optional, List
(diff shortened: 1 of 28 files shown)


shahidee44:
Sphinx wants us to import the modules used for type hinting Q.Q
Looking for an alternative so we don't have to do so.

shahidee44:
@Q-lds Is it possible for us to ignore the warnings and do they cause any potential issues on your end. The warning has something to do with forward references since we don't import the respective modules. It doesn't cause a problem with using the code specifically but sphinx wants us to import the module so it can do its thing.

Q-lds:
I think it may be the time we once and for all fix Sphinx. The general issue is that without some specific import sphinx does not know how to resolve the docs. This makes sense if you think a moment about it.

This may be of relevance https://github.com/tox-dev/sphinx-autodoc-typehints/issues/72; and we'll look into it in the next days

PR #164, opened 2023-01-10
Changed folder structure and access to problems module.

shahidee44:
## Description

- Split the `problem.py` module in the problems folder into seperate python files.
- Added some classes in sub-modules of the problems folder into it's the `__init.py__`. (This would allow users to import certain problems through `openqaoa.problems` without needing to import `openqaoa.problems.sub_module_name` instead)
- Updated unittests that use the problems module
- Updated notebooks that use the problems module
- Removed `from_instance_dict` method from `Problem` Class
- Moved the implementation into a function located in `helper_function` module within the `problems` module
- Moved `check_kwargs` function into `utilities` module

## Checklist

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

test_all.py


examples/community_tutorials/02_docplex_example.ipynb
```diff
@@ -427,7 +427,7 @@
     "ising_encoding = qubo.ising_model  \n",
     "\n",
     "# Print in a df the ising encoding (we need to remove keys: 'problem_instance' and 'metadata')) \n",
-    "ising_encoding_dict = ising_encoding.asdict(exclude_keys=['problem_instance', 'metadata'])\n",
+    "ising_encoding_dict = ising_encoding.asdict(keys_not_to_include=['problem_instance', 'metadata'])\n",
```
Q-lds, line 430: I think this addition is wrong: maybe pull changes from dev? @vishal-ph can confirm

openqaoa/problems/__init__.py
```diff
@@ -0,0 +1,8 @@
+from .knapsack import Knapsack, SlackFreeKnapsack
+from .maximumcut import MaximumCut
+from .minimumvertexcover import MinimumVertexCover
+from .numberpartition import NumberPartition
+from .shortestpath import ShortestPath
+from .tsp import TSP
+from .converters import FromDocplex2IsingModel
+from .qubo import QUBO
```
Q-lds, line 8: Maybe format it all with https://github.com/psf/black

openqaoa/problems/converters.py
```diff
@@ -14,11 +14,11 @@
 
 import numpy as np
 from collections import defaultdict
-from .problem import QUBO
+from .qubo import QUBO
 from typing import Union
 
 
-class FromDocplex2IsingModel:
+class FromDocplex2IsingModel(object):
```
Q-lds, line 21: Is this correct?

openqaoa/problems/minimumvertexcover.py
```diff
@@ -0,0 +1,178 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import networkx as nx
+
+from .helper_functions import check_kwargs
+from .problem import Problem
+from .qubo import QUBO
+
+from collections import Counter
```
Q-lds, line 21: From PEP8 (see https://peps.python.org/pep-0008/#imports)

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.
> Imports should be grouped in the following order:
>  - Standard library imports.
>  - Related third party imports.
>  - Local application/library specific imports.

Let's start by making sure this folder follows the right convention. (Worth checking it black does it automatically)

openqaoa/workflows/optimizer.py
```diff
@@ -19,12 +19,12 @@
...
 import gzip
 
+from openqaoa.problems import QUBO
 from openqaoa.devices import DeviceLocal, DeviceBase
-from openqaoa.problems.problem import QUBO
 from openqaoa.workflows.parameters.qaoa_parameters import CircuitProperties, BackendProperties, ClassicalOptimizer
 from openqaoa.workflows.parameters.rqaoa_parameters import RqaoaParameters
 from openqaoa.qaoa_parameters import Hamiltonian, QAOACircuitParams, create_qaoa_variational_params
-from openqaoa.utilities import get_mixer_hamiltonian, ground_state_hamiltonian, exp_val_hamiltonian_termwise, delete_keys_from_dict, is_valid_uuid, generate_uuid
+from openqaoa.utilities import get_mixer_hamiltonian, ground_state_hamiltonian, exp_val_hamiltonian_termwise, delete_keys_from_dict, is_valid_uuid, generate_uuid, convert2serialize
```
Q-lds, line 27: Do we need to import `convert2serialize` ?

docs/source/problems.rst
```diff
@@ -15,37 +15,43 @@ Problems
 Built-in problems
```
Q-lds, line 15: Maybe we could nest `built-in problems` in the parent key, no? Either way, the current structure is a bit sloppy: let's improve on it! 

image

examples/community_tutorials/02_docplex_example.ipynb
shahidee44: I had an error and saw this as the new attribute. I pulled from dev before this. Ill try again.

openqaoa/problems/converters.py
shahidee44: I usually prefer classes being more verbose about its inheritance.
Its technically only used in python2 and for reverse compatibility. (Object) is automatically added by the interpreter so its unnecessary. Let me know if u think we should remove it. We should also consider it as a future standard for writing objects

Q-lds: If unnecessary maybe I will remove it. @vishal-ph ?

openqaoa/problems/__init__.py
shahidee44: Pushing commit

openqaoa/problems/minimumvertexcover.py
shahidee44: black doesnt do this automatically. I'll fix this.

openqaoa/workflows/optimizer.py
shahidee44: Removed in latest commit. Its not needed.

openqaoa/problems/converters.py
vishal-ph: let's remove it. Yes its a python 2 thing

shahidee44: Although its more cosmetic, we agreed that the difference between the 2 would be if `(object)` is included. The following class would have initialisation as part of its design. If we do not intend to initialise the class, i.e. the class only contains methods, we will not include `(object)`.

openqaoa/utilities.py
```diff
@@ -24,8 +24,8 @@
 import uuid
 import matplotlib.pyplot as plt
 import networkx as nx
-from .qaoa_parameters import Hamiltonian, PauliOp, QAOAVariationalBaseParams
 
+from .qaoa_parameters import Hamiltonian, PauliOp, QAOAVariationalBaseParams
 from .qaoa_parameters.gatemap import *
```
Q-lds, line 29: I'd rather having explicit imports. 
- In the future we may have more stuff in the gatemap and we don't need it here
- It is hard to figure out what function is imported from there if one is just looking at the code

vishal-ph: This is getting all the possible gates in the gatemap file. We do want all gates that are present in the file. If there's extra stuff in future, it may make sense to put that stuff in a separate file other than `gatemap.py`. However, I do see your point. @shahidee44, perhaps we can do the following:
`import gatemap` followed by `gatemap.XYZGateMap` in the code

shahidee44: Fixed

vishal-ph (approved):

PR #166, opened 2023-01-12
Changing data saving workflows methods

raulconchello:
In this PR changes to the dump(s) methods and the way we save the data/files will be implemented.

The header has been already changed, now when we dump the data we find something like:
```
'header'{
 ...,
 'metadata': {'tag1': 'value1',
   'tag2': 'value2',
   'problem_type': 'maximum_cut',
   'key1': 'value1',
   'key2': 'value2',
   'backend_n_shots': 100,
   'optimizer_method': 'nelder-mead',
   'optimizer_jac': 'finite_difference',
   'optimizer_hess': 'finite_difference',
   'circuit_param_type': 'standard',
   'circuit_init_type': 'rand',
   'circuit_p': 2,
   'rqaoa_type': 'adaptive',
   'rqaoa_n_max': 1,
   'rqaoa_n_cutoff': 3}
}
```

### DIFF
--- a/openqaoa/workflows/optimizer.py
+++ b/openqaoa/workflows/optimizer.py
@@ -383,14 +383,17 @@ class Optimizer(ABC):
         else:
             return json.dumps(delete_keys_from_dict(obj= self._serializable_dict(complex_to_string=True), keys_to_delete= exclude_keys), indent=indent)
 
-    def dump(self, file_path:str, indent:int=2, compresslevel:int=0, exclude_keys:List[str]=[]):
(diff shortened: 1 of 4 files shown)


openqaoa/workflows/optimizer.py
```diff
@@ -672,6 +680,14 @@ def _serializable_dict(self, complex_to_string:bool = False):
         # we add the keys of the QAOA object that we want to return
         serializable_dict['data']['input_parameters']['circuit_properties'] = dict(self.circuit_properties)
 
+        # include parameters in the header metadata
+        serializable_dict['header']['metadata']['circuit_param_type'] = serializable_dict['data']['input_parameters']['circuit_properties']['param_type']
```
Q-lds, line 684: Shall we put these into a function? So that the optimiser file doesn't blow up in size?

Q-lds (approved):

shahidee44 (approved):

PR #180, opened 2023-02-06
de_folder_restruct_RQAOA

kidiki:
The way we identify isolated nodes failed for cutoff 1. 
The solution for vanishing instances outputted 2^n states, all degenerate, which is not required.

## Description
-Changing the redefine_problem function to remove the eliminated spins from the difference between registers in the case of isolated nodes. 
- Changing the vanishing instances function to output a single solution only obeying the correlations.

## How Has This Been Tested?
Added: test_total_elimination_cutoff_1_whole_workflow
Modified: test_solution_for_vanishing_instances


src/openqaoa-core/algorithms/rqaoa/rqaoa_utils.py
```diff
@@ -641,47 +619,33 @@ def solution_for_vanishing_instances(hamiltonian: Hamiltonian, spin_map: dict):
...
-        The energy of the first solution wrt the cost Hamiltonian.
+        The energy of the solution wrt the cost Hamiltonian.
     cl_ground_states: `list`
-        List of strings of binary values representing the classical solution of the
-        problem respecting the spin map.
+        List of a string of binary values representing the classical solution of the
+        problem in correspondance with the spin map.
     """
-    cl_ground_states = [""]
+    cl_ground_state = ""
```
Q-lds, line 627: This is not a list of strings

src/openqaoa-core/algorithms/rqaoa/rqaoa_utils.py
```diff
@@ -641,47 +619,33 @@ def solution_for_vanishing_instances(hamiltonian: Hamiltonian, spin_map: dict):
     Returns
     -------
     cl_energy: `float`
-        The energy of the first solution wrt the cost Hamiltonian.
+        The energy of the solution wrt the cost Hamiltonian.
     cl_ground_states: `list`
-        List of strings of binary values representing the classical solution of the
-        problem respecting the spin map.
+        List of a string of binary values representing the classical solution of the
```
Q-lds, line 624: I don't know what a "List of a string" is.

Maybe change to "String representation of binary variables encoding the classical solution .. " or something like that

src/openqaoa-core/algorithms/rqaoa/rqaoa_utils.py
```diff
@@ -641,47 +619,33 @@ def solution_for_vanishing_instances(hamiltonian: Hamiltonian, spin_map: dict):
     Returns
     -------
     cl_energy: `float`
-        The energy of the first solution wrt the cost Hamiltonian.
+        The energy of the solution wrt the cost Hamiltonian.
     cl_ground_states: `list`
```
Q-lds, line 623: I guess the type should be `str`

src/openqaoa-core/algorithms/rqaoa/rqaoa_utils.py
```diff
@@ -641,47 +619,33 @@ def solution_for_vanishing_instances(hamiltonian: Hamiltonian, spin_map: dict):
...
+                # anticorrelated
+                cl_ground_state += str(int(not bool(int(cl_ground_state[parent]))))
 
     # computing the energy of the first one only, assuming degeneracy
-    cl_energy = bitstring_energy(hamiltonian, cl_ground_states[0])
+    cl_energy = bitstring_energy(hamiltonian, cl_ground_state)
+
+    # to match the expected format: list of strings
+    cl_ground_state_list = []
+    cl_ground_state_list.append(cl_ground_state)
```
Q-lds, line 649: why not `cl_ground_state_list = [cl_ground_state]`?

src/openqaoa-core/algorithms/rqaoa/rqaoa_utils.py
kidiki: The energy is float, the ground states are a list of strings usually. Here since it's only one, it's a list of a string.

Q-lds: Oh, you mean that the return is `cl_ground_state_list` and not `cl_ground_states`?

As it stands it is all very confusing

src/openqaoa-core/algorithms/rqaoa/rqaoa_utils.py
kidiki: Changed it directly to cl_ground_state = [cl_ground_state], thanks.

src/openqaoa-core/algorithms/rqaoa/rqaoa_utils.py
kidiki: yeah, true, changed, thanks

vishal-ph (approved):

PR #188, opened 2023-02-22
Dev optimizers notebook

raulconchello:
New Notebook example. It is about the optimizers.

It has no number yet, before merging we should give it a number and change its name.


Q-lds:
This is a very nice notebook! :tada: 

It will also make sense to have something similar on the website :)

vishal-ph:
Very nice notebook, @raulconchello. Congratulations! 

Just one comment, in the section for gradient-free optimizers, I can see `newton-cg`, and `l-bfgs-b` algorithms in the list as well. Be careful, since they require a jacobian for optimization, and therefore, are not gradient-free

raulconchello:
@vishal-ph I have changed what you said.

However, before merging I think we should give it a number.

Q-lds:
How about 13? :D 

Also, is the notebook already featured in `/tests/test_notebooks.py` and in the docs?

raulconchello:
@Q-lds
We can give it the 13, but I feel that the notebooks increase in difficulty and this one is not that technical. 
What do you think, @vishal-ph ?

vishal-ph:
fair point, @raulconchello. Even though it was never a conscious plan to rank the notebooks in increasing order of complexity, it is possible that's how it looks at the moment.

Q-lds:
@vishal-ph and @raulconchello, I think the number of the file is not that relevant. What is important is to organize the notebooks in the docs.

As we already have a lot of them, it probably makes more sense to divide them thematically in the docs and keep a simple increasing counter in the folder

vishal-ph:
sure, then 13 is fine for the moment

vishal-ph (approved):

PR #195, opened 2023-03-02
fix index.rst docs

raulconchello:
code snipset of index.rst updated

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -191,7 +191,7 @@ Then proceed by instantiating the backend device
 
 .. code-block:: python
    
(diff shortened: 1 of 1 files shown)


docs/source/index.rst
```diff
@@ -94,7 +94,7 @@ Workflows can be customised using some convenient setter functions. First, we ne
 
 .. code-block:: python
 
-   from openqaoa.devices import create_device
+   from openqaoa.backends import create_device
```
Q-lds, line 97: The new import is `   from openqaoa import create_device`. But I will fix it directly

raulconchello:
@Q-lds 

I have checked all the code snippets and now all of them are working fine, so if you think nothing else should be changed we can merge it.

Q-lds (approved):

PR #200, opened 2023-03-10
fixed file name dump method workflows

raulconchello:
The dump method was giving the following name:
`{experiment_id}--{atomic_id}`

the correct one is:
`{project_id}--{experiment_id}--{atomic_id}`


Q-lds:
@raulconchello what's the default value for the project id?

Project_id has to be a valid uuid4, it is provided by the user, and it now becomes necessary. That is, if no project id is specified what happens when one tires to save the result?

Q-lds (approved):

PR #208, opened 2023-03-20
UF: Adding the Maximal independent set (MIS) problem

alejomonbar:

Adding the maximal independent set problem class with its corresponding test.
https://en.wikipedia.org/wiki/Maximal_independent_set

## Description

Adding the maximal independent set combinatorial optimization problem to the set of problems. This problem comes with the function `classical_optimization` which uses the docplex solver and `plot_solution` which plots the solution a given solution. 

## Checklist

## Type of change

## How Has This Been Tested?

There is no new test but the test is inside `test_problems`


src/openqaoa-core/problems/maximalindependentset.py
```diff
@@ -0,0 +1,173 @@
...
+
+    Returns
+    -------
+        An instance of the MIS problem.
+    """
+
+    __name__ = "maximal_independent_set"
+
+    def __init__(self, G, penalty:Union[int, float]=2):
+        self.G = G
```
vishal-ph, line 33: should this be `self._G = G`. Otherwise the property and setter method are not utilized.

alejomonbar: I tested the 'self._G = G' but this is the one that does not use the property and setter method. 'self.G = G' works on my local computer to use the property and setter the method. Let me know if you can reproduce it or if I'm making something wrong. @vishal-ph

vishal-ph: yeah, I think you are right. I will resolve the issue

PR #216, opened 2023-03-27
UF: Adding the portfolio optimization problem

alejomonbar:

Adding the [portfolio optimization problem](https://en.wikipedia.org/wiki/Portfolio_optimization) to the set of openqaoa problems. This class method comes with the classical solution method and the plot_solution method to visualize specific solutions.

## Description
New portfolio optimization class in problems.
Test for the problem, random problem generation, classical solution, and plot solution.

## Checklist

## Type of change

## How Has This Been Tested?

./tests/test_problems.py


src/openqaoa-core/problems/portfoliooptimization.py
```diff
@@ -0,0 +1,223 @@
...
+        return PortfolioOptimization(mu, sigma, risk_factor, budget, penalty)
+
+    @property
+    def docplex_model(self):
+        mdl = Model("portfolio_optimization")
+
+        num_assets = self.num_assets
+
+        # Start the docplex model with Model("name of the model")
+        mdl = Model("Portfolio Optimization")
```
vishal-ph, line 100: @alejomonbar, the `mdl` is defined twice. I believe that's a typo. Can you remove one?

alejomonbar: done!

src/openqaoa-core/problems/portfoliooptimization.py
```diff
@@ -0,0 +1,223 @@
...
+
+        if string:
+            solution = ""
+        else:
+            solution = {}
+        for var in cplex_model.iter_binary_vars():
+            if string:
+                solution += str(int(np.round(docplex_sol.get_value(var), 1)))
+            else:
+                solution[var.name] = int(np.round(docplex_sol.get_value(var), 1))
```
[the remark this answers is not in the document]
alejomonbar: round is used because sometimes docplex gives the solution close to 1 but not 1 (in a docplex format), if we want to store this as an integer and use int(value) and the value is 0.999999 it will store a zero. I changed it for just round(docplex_sol.get_value(var)) which gives the same result.

vishal-ph (approved):

PR #220, opened 2023-03-31
SPAMTwirling

kidiki:
## Description
Implementing SPAM Twirling error mitigation technique in OQ. 

## Checklist

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?


src/openqaoa-core/algorithms/baseworkflow.py
```diff
@@ -249,6 +250,26 @@ def set_backend_properties(self, **kwargs):
 
         self.backend_properties = BackendProperties(**kwargs)
         return None
+    
+    @check_compiled
+    def set_error_mitigation_properties(self, **kwargs):
+        """
+        Set the error mitigation properties
+
+        Parameters
```
vishal-ph, line 259: TODO: incomplete docstring

src/openqaoa-core/algorithms/qaoa/qaoa_workflow.py
```diff
@@ -249,11 +250,15 @@ def compile(
...
         )
-
+        
+        backend_dict = self.backend_properties.__dict__.copy()
+        
         self.backend = get_qaoa_backend(
             qaoa_descriptor=self.qaoa_descriptor,
             device=self.device,
-            **self.backend_properties.__dict__,
+            wrapper=SPAMTwirlingWrapper if self.error_mitigation_properties.error_mitigation_technique == 'spam_twirling' else None,
```
vishal-ph, line 259: A suggestion to consider if it is possible. I was looking into the update you made into `get_qaoa_backend`. Let's see if the following approach would work because this might be slightly better code-wise.
```suggestion
self.backend = get_qaoa_backend(
            qaoa_descriptor=self.qaoa_descriptor,
            device=self.device,
            **backed_dict)
if self.error_mitigation_properties.error_mitigation_technique == 'spam_twirling':
     self.backend = SPAMTwirling(self.backend, **wrapper_options)
```

src/openqaoa-core/algorithms/workflow_properties.py
```diff
@@ -262,6 +261,54 @@ def __init__(
...
 
+class ErrorMitigationProperties(WorkflowProperties):
+    """
+    TODO
+    """
+
+    def __init__(
+        self,
+        error_mitigation_technique: Optional[str] = None,
+        n_batches: Optional[int] = 10,
```
vishal-ph, line 272: Two comments here:
- How did we come up with the default value of 10 for `n_batches` 
- If the default value of `n_batches` is 10, you don't need to use `Optional` type hinting.

src/openqaoa-core/utilities.py
```diff
@@ -1427,6 +1484,76 @@ def flip_counts(counts_dictionary: dict) -> dict:
     return output_counts_dictionary
 
 
+def permute_counts_dictionary(
```
vishal-ph, line 1487: This is function is already part of `dev`. Strange that it gets detected as new function. Need to check this

src/openqaoa-pyquil/backends/qaoa_pyquil_qpu.py
```diff
@@ -115,6 +116,8 @@ def __init__(
             cvar_alpha,
         )
         QAOABaseBackendCloud.__init__(self, device)
+        
+        self.gate_applicator = PyquilGateApplicator()
```
vishal-ph, line 120: comment for myself -- I need to convince myself why the `gate_applicator` needs to be an attribute.

src/openqaoa-qiskit/backends/qaoa_qiskit_sim.py
```diff
@@ -106,7 +107,7 @@ def __init__(
             init_hadamard,
             cvar_alpha,
         )
-
+        self.gate_applicator = QiskitGateApplicator()
```
vishal-ph, line 110: same. https://github.com/entropicalabs/openqaoa/pull/220/files#r1164304716

src/openqaoa-core/algorithms/qaoa/qaoa_workflow.py
kidiki: I've edited this now, feel free to resolve this conversation if it's up to standards

src/openqaoa-core/algorithms/workflow_properties.py
kidiki: I think that the number of batches should be of the same order of magnitude as the system size. But it's more or less arbitrary.

src/openqaoa-core/backends/wrapper.py
```diff
@@ -0,0 +1,203 @@
...
+        super().__init__(backend)
+        self.n_batches = n_batches
+        self.calibration_data_location = calibration_data_location
+
+        with open(self.calibration_data_location, "r") as f:
+            calibration_data = json.load(f)
+
+        if (
+            self.calibration_data_location
+            == "calibration_data/8bc00556-2824-4d38-b563-706a803c6401--calibration.json"
```
Q-lds, line 51: Where are these files in the PR?

Q-lds: I am confused by this hardcoded file

kidiki: These lines must be deleted.

src/openqaoa-pyquil/backends/qaoa_pyquil_qpu.py
kidiki: This is necessary because the wrapper needs to have access to the gate_applicator for the specific backend so that it can generate an empty circuit for the backend.

tests/test_qpu_braket.py
```diff
@@ -116,6 +116,7 @@ def test_circuit_angle_assignment_qpu_backend_w_hadamard(self):
             qaoa_descriptor, aws_device, shots, None, None, True, 1.0
         )
         qpu_circuit = aws_backend.qaoa_circuit(variate_params)
+
```
vishal-ph, line 119: are the extra newlines coming out of automated formatting?

tests/test_utilities.py
```diff
@@ -923,6 +910,169 @@ def test_exp_val_hamiltonian_termwise(self):
                 corr_matrix[j], num_corr_matrix[j]
             ), f"Computed correlation matrix is incorrect"
 
+    def test_calculate_calibration_factors(self):
+        """
+        Tests the function that computes the calibration factors.
+
+        The test consists in ...
```
vishal-ph, line 917: [tag] To be completed

src/openqaoa-braket/backends/gates_braket.py
```diff
@@ -45,49 +48,38 @@ def apply_2q_rotation_gate(
...
+            if hasattr(gate, "rotation_object"):
                 # *args must be of the following format -- (qubit_1,qubit_2,rotation_object,circuit)
                 return self.apply_2q_rotation_gate(selected_braket_gate, *args)
             else:
                 # *args must be of the following format -- (qubit_1,qubit_2,circuit)
                 return self.apply_2q_fixed_gate(selected_braket_gate, *args)
         else:
-            raise ValueError("Only 1 and 2-qubit gates are supported.")
\ No newline at end of file
+            raise ValueError("Only 1 and 2-qubit gates are supported.")
```
Q-lds, line 85: Maybe it makes sense to be more explicit (the code fail also when n_qubits is none):
```
raise ValueError(f"Only 1 and 2-qubit gates are supported. Instead, I received n_qubits={n_qubits}")
```

src/openqaoa-braket/backends/qaoa_braket_qpu.py
```diff
@@ -132,6 +133,13 @@ def qaoa_circuit(self, params: QAOAVariationalBaseParams) -> Circuit:
         qaoa_circuit: `Circuit`
             The final QAOA circuit constructed using the angles from variational params.
         """
+        parametric_circuit = deepcopy(self.parametric_circuit)
+
+        if self.append_state:
+            parametric_circuit += self.append_state
+
+        # TODO: needs to be fixed --> measurement operations on problem qubits
+        parametric_circuit += Probability.probability()
```
Q-lds, line 142: What is the role of this?

src/openqaoa-core/algorithms/workflow_properties.py
```diff
@@ -262,6 +261,62 @@ def __init__(
...
+        The path to the file containing calibration data for the specific device.
+    """
+
+    def __init__(
+        self,
+        error_mitigation_technique = None,
+        n_batches = 10,
+        calibration_data_location = None,
+    ):
+        self.error_mitigation_technique = error_mitigation_technique
```
Q-lds, line 284: Always cast the user input to lower string. Otherwise, the code Breaks if I say `error_mitigation_technique='SPAM_twirling'`

tests/test_qpu_braket.py
kidiki: I'll format all relevant files with black once again before merging.

src/openqaoa-core/backends/wrapper.py
```diff
@@ -0,0 +1,200 @@
...
+    def get_counts(self, params, n_shots=None, seed=42):
+        """
+        Overrides the get_counts function of the backend object.
+        Modified function to...
+            divide into batches
+            change the self.append_state according to the schedule, s.
+            get the counts and classically negate them
+            combine all batches into a count dict under BFA
+        """
+        random.seed(seed)
```
Q-lds, line 99: This needs to be a bit documented. Also, I am not sure whether fixing the seed all of a sudden will have unforeseen side effects across the code.

I am not saying that this is wrong, but it may be dangerous!

Q-lds: See https://stackoverflow.com/questions/52544935/best-way-to-revert-to-a-random-seed-after-temporarily-fixing-it

src/openqaoa-core/backends/wrapper.py
```diff
@@ -0,0 +1,200 @@
...
+        # list of integers whose binary representation signifies which qubits to be flipped at every batch
+        s_list = []
+        for _ in range(0, self.n_batches):
+            s_list.append(random.getrandbits(self.backend.n_qubits))
+
+        n_shots = self.backend.n_shots if n_shots == None else n_shots
```
Q-lds, line 105: How is it possible for n_shots to ne none here?

src/openqaoa-core/backends/wrapper.py
```diff
@@ -0,0 +1,200 @@
...
+        self.measurement_outcomes = counts
+
+        return counts
+
+    @round_value
+    def expectation_value_spam_twirled(
+        self, counts: Dict, hamiltonian: Hamiltonian, calibration_factors: dict
+    ):
+        """
+        TODO
```
Q-lds, line 153: TODO!

src/openqaoa-core/backends/wrapper.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+
+        terms = [term.qubit_indices for term in hamiltonian.terms]
+        hamiltonian_as_dict = hamiltonian.hamiltonian_dict()
```
Q-lds, line 157: No need to create an extra variable. Just pass `hamiltonian.hamiltonian_dict()` as the argument in lines 169 and 177

src/openqaoa-core/utilities.py
```diff
@@ -1159,6 +1141,90 @@ def exp_val_hamiltonian_termwise(
...
+                calibration_registers_dict[j_phys],
+            )  # calibration indices, i.e. to which location on the measurement string each physical qubit corresponds to, ex: (63, 61)
+            exp_val_zz = exp_val_pair((i_cal, j_cal), calibration_measurements)
+
+            calibration_factors.update(
+                {(i, j): exp_val_zz}
+            )  # calibration factors are calculated for the terms in the hamiltonian/problem
+
+        # If constant term, ignore
+        if len(term) == 0:
```
Q-lds, line 1214: Should this if be on its own? Or is it part of the previous conditional (and it should be an else or something)?

src/openqaoa-core/utilities.py
```diff
@@ -1427,6 +1488,76 @@ def flip_counts(counts_dictionary: dict) -> dict:
...
+                [basis_state[mapping[i]] for i in range(len(basis_state))]
+            )
+            return permuted_string
+
+        permuted_counts.update({permuted_string: counts})
+
+    return permuted_counts
+
+
+def negate_counts_dictionary(counts_dictionary: dict, s: int) -> dict:
```
Q-lds, line 1533: Is the negation just a bitflip? If so maybe bitflip is easier to read than negate?

src/openqaoa-core/algorithms/qaoa/qaoa_workflow.py
```diff
@@ -250,12 +251,24 @@ def compile(
...
+        backend_dict = self.backend_properties.__dict__.copy()
+
         self.backend = get_qaoa_backend(
             qaoa_descriptor=self.qaoa_descriptor,
             device=self.device,
-            **self.backend_properties.__dict__,
+            **backend_dict,
         )
 
+        if (
```
vishal-ph, line 262: Let's add some comments here for future reference.

tests/qpu_calibration_data/spam_twirling_mock.json
```diff
@@ -0,0 +1,19 @@
+{
```
Q-lds, line 1: Maybe add a comment describing what is in the file? And how it is used?

Looking at it I don't really know what the register is!

tests/test_wrapper.py
```diff
@@ -0,0 +1,242 @@
...
+            calibration_data_location=self.calibration_data_location,
+        )
+
+        self.wrapped_obj_trivial = SPAMTwirlingWrapper(
+            self.qiskit_shot_backend,
+            n_batches=1,
+            calibration_data_location=self.calibration_data_location,
+        )
+
+    def test_wrap_any_backend(self):
```
Q-lds, line 93: Tag this tests as qvm

Q-lds: Wait, this also requires a QPU access

src/openqaoa-braket/backends/qaoa_braket_qpu.py
vishal-ph: (for me)

src/openqaoa-qiskit/backends/qaoa_qiskit_qpu.py
```diff
@@ -128,24 +131,38 @@ def qaoa_circuit(self, params: QAOAVariationalBaseParams) -> QuantumCircuit:
         qaoa_circuit: `QuantumCircuit`
             The final QAOA circuit after binding angles from variational parameters.
         """
+        parametric_circuit = deepcopy(self.parametric_circuit)
```
vishal-ph, line 134: This may be the part slowing down the code, perhaps.

kidiki: maybe, but we do need this one since we don't want to modify the original, already-constructed, self.parametric_circuit, right?

src/openqaoa-core/backends/wrapper.py
kidiki: For now, it's been changed to. 

if seed is not None:
            random.seed(seed)

Have to decide if we want to create a new instance, and if this should be done not only here but everywhere we set a seed.

src/openqaoa-core/backends/wrapper.py
kidiki: Are you sure that this won't slow the code? Because it will be calling this function often, it's inside a for loop over all terms of the hamiltonian.

src/openqaoa-core/utilities.py
kidiki: changed

tests/qpu_calibration_data/spam_twirling_mock.json
kidiki: added a README in the calibration data folder

src/openqaoa-braket/backends/gates_braket.py
kidiki: Sure, but that's not directly relevant for this PR.

src/openqaoa-core/algorithms/qaoa/qaoa_workflow.py
kidiki: Done. Please review.

tests/test_utilities.py
kidiki: Completed. However, should I mode the comments from the code to the doc strings?

src/openqaoa-core/backends/wrapper.py
Q-lds: Good point: try!

Time it and report which one is faster. You can use https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-timeit

src/openqaoa-core/algorithms/workflow_properties.py
vishal-ph: Let's add a comment to justify this default value. In the docstring itself should work

src/openqaoa-qiskit/backends/qaoa_qiskit_qpu.py
vishal-ph: we construct a new circuit after `bind_parameters`, why can't we append these instructions to this circuit instead of deepcopying the parametric circuit and then adding instructions to it? After this step I mean `circuit_with_angles = self.parametric_circuit.bind_parameters(memory_map)`

src/openqaoa-qiskit/backends/qaoa_qiskit_qpu.py
```diff
@@ -128,24 +131,38 @@ def qaoa_circuit(self, params: QAOAVariationalBaseParams) -> QuantumCircuit:
...
+        else:
+            for idx, qubit in enumerate(self.final_mapping[0 : len(self.problem_reg)]):
+                cbit = self.creg[idx]
+                parametric_circuit.measure(qubit, cbit)
+                
 
         angles_list = self.obtain_angles_for_pauli_list(self.abstract_circuit, params)
         memory_map = dict(zip(self.qiskit_parameter_list, angles_list))
-        circuit_with_angles = self.parametric_circuit.bind_parameters(memory_map)
+        new_parametric_circuit = parametric_circuit.bind_parameters(memory_map)
```
vishal-ph, line 150: also, the output of this function is not a parametric circuit, so why are calling it `new_parametric_circuit` as opposed to the older name?

src/openqaoa-core/algorithms/workflow_properties.py
kidiki: Done. Please review

Q-lds: Sorry, I don't see it!

I'd change the code to

```
    def __init__(
        self,
        error_mitigation_technique: str = None,
        n_batches:int = 10,
        calibration_data_location:Optional[str] = None,
    ):
        self.error_mitigation_technique = error_mitigation_technique.lower()
```

src/openqaoa-braket/backends/gates_braket.py
Q-lds: Isn't this new code?

src/openqaoa-core/algorithms/workflow_properties.py
```diff
@@ -262,6 +261,71 @@ def __init__(
...
+        The path to the file containing calibration data for the specific device.
+    """
+
+    def __init__(
+        self,
+        error_mitigation_technique=None,
+        n_batches=10,
+        calibration_data_location=None,
+    ):
+        self.error_mitigation_technique = (
```
kidiki, line 284: @Q-lds this is the lower case

Q-lds: Got it, thanks!

I would still force the datatype to be a string in the `init`, and then just do `.lower()`

src/openqaoa-pyquil/backends/gates_pyquil.py
```diff
@@ -6,35 +6,39 @@
...
 
-    library = 'pyquil'
+    library = "pyquil"
+
+    def create_quantum_circuit(self, n_qubits) -> Program:
+        """
+        Function which creates and empty circuit for the specific backend.
+        Needed for twirling but can be used inside parametric circuit too
+        instead of creating parametric_circuit = QuantumCircuit(self.qureg)
+        TODO this function doesn't require n_qubits but the one for qiskit does.
```
kidiki, line 29: this function doesn't require n_qubits but the one for qiskit does. Is there a way to make it more generic or we just pass an extra param?

src/openqaoa-braket/backends/qaoa_braket_qpu.py
```diff
@@ -143,8 +151,10 @@ def qaoa_circuit(self, params: QAOAVariationalBaseParams) -> Circuit:
                 angles_list,
             )
         )
-        new_parametric_circuit = self.parametric_circuit.make_bound_circuit(memory_map)
-        return new_parametric_circuit
+        circuit_with_angles = parametric_circuit.make_bound_circuit(memory_map)
+
+        self.append_state = []
```
vishal-ph, line 156: update this

src/openqaoa-braket/backends/gates_braket.py
vishal-ph: we can fix this separately in another PR.

src/openqaoa-core/algorithms/workflow_properties.py
kidiki: okay, changed, thanks

vishal-ph (approved):

PR #224, opened 2023-04-07
UF: Adding the vehicle routing problem (VRP)

alejomonbar:

Adding the vehicle routing problem class with its corresponding test.
https://en.wikipedia.org/wiki/Vehicle_routing_problem
## Description

Adding the vehicle routing problem combinatorial optimization problem to the set of problems. This problem comes with the function 'classical_optimization' which uses the 'docplex' solver and plot_solution which plots the 'networkx' graph for a given solution.

## Checklist

## Type of change

## How Has This Been Tested?

the test is inside 'test_problems'


src/openqaoa-core/problems/vehiclerouting.py
```diff
@@ -0,0 +1,380 @@
...
+                For the unbalaced method refer to https://arxiv.org/abs/2211.13914
+                For the slack method: https://en.wikipedia.org/wiki/Slack_variable
+            subtours: list[list]
+                Manually add the subtours to be avoided
+            seed: int
+                Seed for the random problems.
+
+        Returns
+        -------
+            A random instance of the maximal independent set problem.
```
vishal-ph, line 116: Typo in the text, it refers to MIS instead of VSP

src/openqaoa-core/problems/vehiclerouting.py
```diff
@@ -0,0 +1,380 @@
...
+        for i in range(n_nodes - 1):
+            for j in range(i + 1, n_nodes):
+                r = np.sqrt((pos[i][0] - pos[j][0]) ** 2 + (pos[i][1] - pos[j][1]) ** 2)
+                G.add_weighted_edges_from([(i, j, r)])
+        return VRP(
+            G, pos, n_vehicles, subtours=subtours, method=method, penalty=penalty
+        )
+
+    @property
+    def docplex_model(self):
```
vishal-ph, line 143: Can we add a docstring to this method?

src/openqaoa-core/problems/vehiclerouting.py
```diff
@@ -0,0 +1,380 @@
...
+
+    __name__ = "vehicle_routing"
+
+    def __init__(
+        self,
+        G,
+        pos,
+        n_vehicles,
+        depot: int = 0,
+        subtours: list = [],
```
vishal-ph, line 55: I think its probably a better idea to pass `None` as default argument for subtours, as `subtours: Optional[list] = None`, then set it to empty list inside the function body. 
`self.subtours = [] if subtours is None else subtours`

src/openqaoa-core/problems/vehiclerouting.py
alejomonbar: done!

src/openqaoa-core/problems/vehiclerouting.py
alejomonbar: ok

src/openqaoa-core/problems/vehiclerouting.py
alejomonbar: sure!

src/openqaoa-core/problems/vehiclerouting.py
vishal-ph: Typically, it's a bad practice to pass mutable default arguments. In certain situations, they may have silent side effects.  For e.g., you can look at this https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

src/openqaoa-core/problems/vehiclerouting.py
vishal-ph: thanks

src/openqaoa-core/problems/vehiclerouting.py
```diff
@@ -0,0 +1,425 @@
...
+
+    __name__ = "vehicle_routing"
+
+    def __init__(
+        self,
+        G,
+        pos,
+        n_vehicles,
+        depot: int = 0,
+        subtours: list = -1,
```
vishal-ph, line 56: Did the default = `None` not work for this argument?

alejomonbar: Yes, but I want for the default argument to do all the subtours and subtours=[] for not adding any subtour constraint. subtours = None has the problem if one wants to convert the qubo into a dict using `todict` because it does not save the key if the value is None. (I just found it when testing).

vishal-ph (approved):

src/openqaoa-core/problems/vehiclerouting.py
alejomonbar: I see! Thanks

PR #230, opened 2023-05-03
UF: Adding the traveling salesman problem (TSP) linear programming version 

alejomonbar:

Adding the traveling salesman problem class with its corresponding test.
https://en.wikipedia.org/wiki/Travelling_salesman_problem

## Description

Adding the traveling salesman problem combinatorial optimization problem to the set of problems. This problem comes with the function 'classical_optimization' which uses the 'docplex' solver, plot_solution which plots the 'networkx' graph for a given solution, and a `get_distance` function to get the total distance traveled from a given solution.

Additionally, I added a subfolder to organize the tests for the problems.

## Checklist

## Type of change

## How Has This Been Tested?

the test is inside './tests/test_problems/test_tsp.py'


src/openqaoa-core/problems/tsp.py
```diff
@@ -350,3 +351,111 @@ def qubo(self):
         # Convert to Ising equivalent since variables are in {0, 1} rather than {-1, 1}
         ising_terms, ising_weights = QUBO.convert_qubo_to_ising(n, terms, weights)
         return QUBO(n, ising_terms, ising_weights, self.problem_instance)
+
+class TSP_LP(VRP):
```
shahidee44, line 355: Is there a reason why `TSP_LP` does not inherit for `TSP` object class instead?

alejomonbar: Because the linear programming version of the TSP which is the TSP_LP is a specific case of the VRP, or the VRP with just one vehicle. @shahidee44

shahidee44: Hmmm I see. Do you think it makes sense for us to rename the class to something like `VRP_LP` instead?
But I guess if the nature of the problem classes are similar this current naming is fine.

Let's include this information in the docstring. i.e. TSP_LP is a specific case of the VRP/its the VRP with just one vehicle.

alejomonbar: I added a note indicating that the TSP_LP can be seen as a particular case of the VRP for just one vehicle. @shahidee44

shahidee44 (approved):

PR #236, opened 2023-05-23
Added Scripts for Deployment Testing to TestPypi

shahidee44:
## Description

- Fixed metapackage deployment issue.
- Added/Updated scripts for deployment testing of individual plugins to TestPypi
- Updated scripts for local install and testing for individual plugins
- Versions are now caveated with `-rcX` to represent pre-release. (These versions can only be downloaded using the `--pre` flag)
- `openqaoa` metapackage now installs its dependencies from Pypi.
- **Fixed `test_dev.yml` workflow and `Dockerfile` to use new Makefile commands.**

## Checklist

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## How Has This Been Tested?


shahidee44:
`openqaoa` metapackage renamed to `openqaoa-meta` for testing purposes. (To change back before merging with `dev`)

.github/workflows/test_dev.yml
```diff
@@ -34,8 +34,7 @@ jobs:
           python -m venv env
           source env/bin/activate
           python -m pip install --upgrade pip
-          pip install .
-          pip install -e ./src/openqaoa-core[tests]
+          make dev-install-tests
           pip install ipykernel
```
vishal-ph, line 38: This is perhaps a bit of nitpicking, but I believe ipykernel is already installed as its a dependency of dev-install-tests, no?

Dockerfile
```diff
@@ -5,4 +5,4 @@ RUN mkdir -p /openqaoa
 
 ADD ./ /openqaoa/
 
-RUN pip3 install /openqaoa/.
+RUN cd openqaoa && make dev-install-tests
```
vishal-ph, line 8: do we need to install the `test` version of openqaoa from the Dockerfile? It doesn't seem like we were installing the `test` version before

scripts/install_and_test_openqaoa_azure.sh
```diff
@@ -16,8 +16,4 @@ for entry in "${modulesList[@]}"; do
     cd src/$entry
     pytest -n auto tests
     cd "../.."
-done
```
vishal-ph, line 19: Can you help me understand why we are not uninstalling packages anymore after testing?

scripts/install_and_test_openqaoa_dev.sh
```diff
@@ -1,9 +1,6 @@
...
 
-# TODO: Dynamically generate names (The order of installing is important here.)
-modulesList=("src/openqaoa-core" "src/openqaoa-qiskit" "src/openqaoa-braket" "src/openqaoa-pyquil" "src/openqaoa-azure")
-
 pip install .
-pytest tests/ src/*/tests
-pip uninstall -y openqaoa
\ No newline at end of file
+pytest tests/ src/*/tests -n auto
+pip uninstall -y openqaoa-meta
```
vishal-ph, line 6: I believe the `meta` in the name needs to go

.github/workflows/test_dev.yml
shahidee44: I don't think we install `ipykernel` for `dev-install-tests`.
https://github.com/entropicalabs/openqaoa/blob/dev_test_pypi/src/openqaoa-core/requirements_test.txt

It might be installed as a dependency for something else maybe. Should we add it in?

Dockerfile
shahidee44: Good point. I'll try the normal install and see if tests work.

scripts/install_and_test_openqaoa_dev.sh
shahidee44: Good catch. Removed `meta` reference.

scripts/install_and_test_openqaoa_azure.sh
shahidee44: I think in the end, the only use-case for the scripts are for situations where the user wants to install the modules with the relevant unittests and run the tests. 

The uninstallation became double work for the script that was trying to test the modules by running the scripts, since the packages got uninstalled and reinstalled. We can add these lines back but i felt like these scripts were more useful if the modules were not uninstalled at the end of it, barring also the time we save when we run [run_all_install_and_test.sh](https://github.com/entropicalabs/openqaoa/pull/236/files#diff-f9dffec80028b1c574b13e49e3365d8b489a64c8dca70d27d14147efc1ee599b)

.github/workflows/test_dev.yml
vishal-ph: fair fair, sorry, I may have confused it with ipython

vishal-ph (approved):

Issue #234, opened 2023-05-17
Expose brute force solver through OpenQAOA workflows


vishal-ph:
### Prerequisites
Before raising this feature request, I made sure:

- [x] It does not lead to duplication.
- [x] It has not already been requested.

### Is your feature request related to a problem? Please describe.
OpenQAOA has a function called `ground_state_hamiltonian` that takes as input the Cost Hamiltonian corresponding to the QUBO problem and outputs the degenerate solutions (or ground states of the Hamiltonian) along with the lowest energy (i.e. ground state energy). The function is defined in `src/openqaoa-core/openqaoa/utilities.py`.

Expose this function as a `QAOA` workflow class method so users can easily access it. 

### Describe the solution you'd like
- Implement the method that calls this function from within the QAOA workflow class
- Make sure to impose an artificial limit on the size of problem (in terms of the number of qubits) to prevent memory overflow.

Ideal workflow
1. `q = QAOA()`
2. `q.compile(qubo)`
3. `q.solve_brute_force()`

### Describe alternatives you've considered
The existing way to obtain the brute force solution of a qubo is described as follows:
1. construct the problem `qubo` 
2. `from openqaoa.utilities import ground_state_hamiltonian
3. Construct the hamiltonian object by calling `prob_hamiltonian = qubo.hamiltonian`
4. Obtain the solution `solution, energy = ground_state_hamiltonian(prob_hamiltonian)`

lumapools:
Hi, I have attempted to implement the `brute_force` method, could you please assign me to this issue? I am just having issues writing a test (in `test_workflows.py`) for the method because the imported modules are from pip. So I am not sure if I should include a test case or not.

vishal-ph:
@lumapools, just did! Thanks

lumapools:
Thank you! Just created the PR with the feature and new tests to be added

lumapools:
I think it is a good idea too! How should we go about merging ? Should we first merge @GiacomoFrn 's PR and then I modify my PR (or close it to create a new one) to only include the (new) tests for @GiacomoFrn code ?

Q-lds:
Perfect! 

@vishal-ph I think we follow @lumapools suggestion. What do you think?

vishal-ph:
thanks @lumapools, @GiacomoFrn for contributing to OpenQAOA and helping solve this issue!

GiacomoFrn:
[...]

PR #239, opened 2023-05-26
Fixing issue #234: Brute force solver as OpenQAOA workflow class method

GiacomoFrn:
## Description

- Now  the QAOA workflow class contains the method `solve_brute_force()` that calls the `ground_state_hamiltonian` function from `openqaoa.utilities`. 
- Results are stored as a dictionary in `self.brute_force_results` and can be printed out with the `verbose` flag.
- The parameter `bounded` imposes the artificial limit of 25 qubits to prevent memory overflow
- **Fixes # 234**

### DIFF
--- a/src/openqaoa-core/algorithms/qaoa/qaoa_workflow.py
+++ b/src/openqaoa-core/algorithms/qaoa/qaoa_workflow.py
@@ -319,7 +319,7 @@ class QAOA(Workflow):
         """
 
         if self.compiled is False:
-            raise ValueError("Please compile the QAOA before optimizing it!")
+            raise ValueError("Please compile the QAOA before running the brute force solver!")
         
         # compute the exact ground state and ground state energy of the cost hamiltonian
         energy, configuration = ground_state_hamiltonian(self.cost_hamil, bounded=bounded)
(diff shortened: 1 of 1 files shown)


vishal-ph:
@GiacomoFrn, thank you for your PR solving issue #234, please give us a few days while we review the code.

src/openqaoa-core/algorithms/qaoa/qaoa_workflow.py
```diff
@@ -19,6 +19,7 @@
 )
 from ...utilities import get_mixer_hamiltonian, generate_timestamp
 from ...optimizers.qaoa_optimizer import get_optimizer
+from ...utilities import ground_state_hamiltonian
```
vishal-ph, line 22: can you add the import for the function directly to line 20?

src/openqaoa-core/algorithms/qaoa/qaoa_workflow.py
```diff
@@ -301,6 +302,38 @@ def compile(
...
+            If set to True, the function will not perform computations for qubit
+            numbers above 25. If False, the user can specify any number. Defaults
+            to True.
+        verbose: `bool`, optional
+            If set to True, the function will print the results of the computation.
+            Defaults to False.
+        """
+
+        if self.compiled is False:
+            raise ValueError("Please compile the QAOA before optimizing it!")
```
vishal-ph, line 323: The error message should be different

GiacomoFrn:
@vishal-ph, I should have fixed the issues you raised

Q-lds:
Some of the tests are failing because they required some secretes that are not available to forks! 

I forgot that was the case! :)

I will merge it now, and then let @lumapools add the tests

Issue #233, opened 2023-05-17
Prevent double authentication while using `openqaoa-azure`


vishal-ph:
### Prerequisites
Before raising this issue, I have already checked that I am:

- [x] running the latest version
- [x] made sure that this issue has not already been filed

### Describe the bug
Running a QAOA problem using the OpenQAOA workflow as described in [this](https://github.com/entropicalabs/openqaoa/blob/main/examples/12_testing_azure.ipynb) notebook will require web authentication twice before executing the run. For reference, see the attached screenshot.

This happens because DeviceAzure, which is defined in `src/openqaoa-azure/openqaoa_azure/backends/devices.py`, has a method called `check_connection`, that is called twice, once from within the workflow and the other time from within the backend object. 

### To Reproduce
1. Create the Azure device object
2. Create a `qubo` for a problem of choice
3. Initialize the QAOA workflow as `q = QAOA()`
4. Call the compile method and pass the QUBO via `q.compile(qubo)`

### Expected behavior 
Upon executing the above steps, the web authenticator should only prompt once for device authentication

### Screenshots
![image](https://github.com/entropicalabs/openqaoa/assets/35625965/98a1e761-cbb5-46c1-9769-bd5920399f0f)

WingCode:
@vishal-ph I would like to take a stab at this issue. Could you assign it to me?

vishal-ph:
NOTE: - Please branch out from openqaoa/dev and will make the PR into openqaoa/dev. Thanks

vishal-ph:
@TerraVenil, thanks for taking a look at this issue! Regarding your questions:
- Judging from your screenshot, it seems your CLI is not set up with Azure credentials. To access the devices from Azure and use them with OpenQAOA, you must first set up Azure CLI. Instructions to do that can be found [here](https://openqaoa.entropicalabs.com/devices/azure-quantum/#how-to-connect-to-azure-quantum). In case, the issue is something else, please let me know!
- For the second part of your question, you can look [here](https://github.com/entropicalabs/openqaoa/blob/137d3f0a10391adce5bb23cf7536f2d221defb79/src/openqaoa-azure/backends/devices.py#LL59C59-L59C59) and notice that the `check_connection` method of the device object verifies if connections to both `provider` and `QPU` are made, and returns `True` if both `self.provider_connected` and `self.qpu_connected` are True, or if `self.provider_connected` is True and `self.device_name == ""`
[...]

PR #249, opened 2023-06-08
Prevent double authentication while using openqaoa-azure

TerraVenil:
## Description

This PR is for solving issue https://github.com/entropicalabs/openqaoa/issues/233.

## Checklist

## Type of change

## How Has This Been Tested?

By running tests locally for openqaoa-azure and openqaoa-core.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,8 +1,8 @@
 [pytest]
 markers =
     qpu: marks tests that require a QPU connection (deselect with '-m "not qpu"')
-    api: marks tests that require a coonection to a QC cloud (deselect with '-m "not api"')
-    qvm: marks tests that require an attive Rigetti QVM and QILC compiler (deselect with '-m "not qvm"')
+    api: marks tests that require a connection to a QC cloud (deselect with '-m "not api"')
+    qvm: marks tests that require an active Rigetti QVM and QILC compiler (deselect with '-m "not qvm"')
     docker_aws: marks tests that require to build aws docker (deselect with '-m "not docker_aws"')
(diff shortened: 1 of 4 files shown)


shahidee44 (approved):
Solution looks good. 

Additional Comments:

- The proposed solution only accounts for the case where the user has tried to run `check_connection` on the same device before the device is being passed into the `Backend`
- It does not account for any other case, for example, cases where there are incorrect credentials or names. In those cases, the `check_connection` method is run again for the same device.

Leaving this comment here so we have reference. Approving changes.

shahidee44:
@TerraVenil You will need to update this PR with the latest version of `dev` so that it can pass the relevant unittests. :)

shahidee44:
Thats a good point on Authentication Errors. The intention of the Device object is to catch `Exception` (s) that occur during the authentication process and just return a boolean based on the success of the authentication. The handling of the error would be done by other layers controlling the Device object. This was done to ensure consistency in error handling across the different Devices.

Also, it seems like I'm made a mistake in my initial comment, the new amendment only authenticates iff the Device object has not tried to authenticate before. The booleans `provider_connected` and `qpu_connected` change depending on whether certain parts of the authentication process is successful.

It is then more appropriate to note that:

- If a Device Object has already tried to be authenticated before passing through to the Backend. Then, there will be no attempt to reauthenticate.

This is an important point to note since in some fringe case where the user authenticates, initializes the Device and runs `check_connection`, makes some change to the internal attributes to the Device after the fact, then passes that modified Device into the Backend. The Device object will not be reauthenticated. 

The main point here would be that we need to let users know that the Device object would need to be recreated or `check_connection` has to be run again on the same object, if the user is looking to reauthenticate the Device.

Issue #235, opened 2023-05-17
Add result statistics on measurement outcomes in QAOAResult object


vishal-ph:
### Prerequisites

- [x] It does not lead to duplication.
- [x] It has not already been requested.

### Is your feature request related to a problem? Please describe.

The QAOAResult object stores information regarding the QAOA optimization routine. This result object can be accessed post optimization through the `QAOA` workflow object via the following steps:
1. Define the problem `qubo`
2. Initialize QAOA workflow object `q = QAOA()`
3. Set QAOA properties via setter methods
4. Compile the object `q.compile(qubo)`
5. Run the optimization `q.optimize()`
6. Obtain the results of optimization as `qaoa_result = q.result`

Add a method under the `QAOAResult` class that computes some statistical quantities for the measurement outcomes.

### Describe the solution you'd like
The QAOA object can store measurement outcomes (i.e. counts) from each circuit evaluation in the `QAOAResult` object.

- The method sorts the measurement outcomes (a dictionary) in the descending order of counts. 
- It computes the `mean` and `standard deviation` for the optimized set of measurement outcomes. 
- The function also optionally computes these quantities for all intermediate measurement outcomes during optimization if the user specifies. 
- The method should raise an error if the user requires statistics on all intermediate measurement outcomes but hasn't specified saving them during optimization.

### Additional context
An approach adding statistical tools to understand QAOA results better

WingCode:
@vishal-ph I would like to take a stab at this issue. Could you assign it to me?

vishal-ph:
@WingCode, sure! Thanks!

Q-lds:
@TerraVenil, thanks for your contribution! :)

PR #252, opened 2023-06-10
Add result statistics on measurement outcomes in QAOAResult object

TerraVenil:
## Description

This PR is for solving issue https://github.com/entropicalabs/openqaoa/issues/235.

## Checklist

## Type of change

- [x] This change requires a documentation update

## How Has This Been Tested?

test_qaoa_results_calculate_statistics_full, test_qaoa_results_calculate_statistics_without_intermediate, test_qaoa_results_calculate_statistics_raise_value_error.

### DIFF
--- a/src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_result.py
+++ b/src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_result.py
@@ -629,3 +629,56 @@ class QAOAResult:
             ],
         }
         return best_results
+    
+    def calculate_statistics(self, include_intermediate=False) -> dict:
+        """
+        A function to calculate statistics of measurement outcomes associated with a QAOA workflow
+
+        Parameters
+        ----------
+        include_intermediate: `bool`
+            Whether it is necessary to calculate statistics for intermediate results. Defaults to False.
+        """
+        if len(self.intermediate["measurement_outcomes"]) == 0 and include_intermediate == True:
+            raise ValueError(
+                "Please specify saving intermediate measurements during optimization."
+            )
+
+        if isinstance(self.optimized["measurement_outcomes"], dict):
+            optimized_measurement_outcomes = self.optimized["measurement_outcomes"]
+        elif isinstance(self.optimized["measurement_outcomes"], np.ndarray):
+            optimized_measurement_outcomes = self.get_counts(
+                self.optimized["measurement_outcomes"]
+            )
+        else:
+            raise TypeError(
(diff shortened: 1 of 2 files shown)


src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_result.py
```diff
@@ -629,3 +629,56 @@ def lowest_cost_bitstrings(self, n_bitstrings: int = 1) -> dict:
...
+        A function to calculate statistics of measurement outcomes associated with a QAOA workflow
+
+        Parameters
+        ----------
+        include_intermediate: `bool`
+            Whether it is necessary to calculate statistics for intermediate results. Defaults to False.
+        """
+        if len(self.intermediate["measurement_outcomes"]) == 0 and include_intermediate == True:
+            raise ValueError(
+                "Please specify saving intermediate measurements during optimization."
```
Q-lds, line 644: Maybe we can be more explicit. How about:
 
"_The underlying QAOA object does not seem to have any intermediate measurement result. Please, consider saving intermediate measurements during optimization by setting `optimization_progress=True` in your workflow._"

Or a less verbose version of it: the important thing (I think) is to give the end user a pointer to what they need to do in order to use this feature :)

src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_result.py
```diff
@@ -629,3 +629,56 @@ def lowest_cost_bitstrings(self, n_bitstrings: int = 1) -> dict:
...
+                    )
+        else:
+            raise TypeError(
+                f"The measurement outcome {type(self.intermediate['measurement_outcomes'])} is not valid."
+            )
+
+        def sorted_mean_std_deviation(counts: dict):
+            values = list(counts.values())
+            return {
+                'sorted': dict(sorted(counts.items(), key=lambda x: x[1], reverse=True)),
```
Q-lds, line 676: I am having second thoughts about this, perhaps we can remove it?

I am thinking it may not be very informative after all, and the user can always compute it if they so wish.

On the other hand, maybe we could return also mean and std for some percentiles? What do you think?

Q-lds:
@TerraVenil, very nice contribution! I just left two minor comments/suggestions and then we are good to merge :)

src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_result.py
Q-lds: Actually, no need to implement this now. I think it is fine the way it is :)

Q-lds (approved):

PR #262, opened 2023-07-06
Dev

vishal-ph:
## Description

- Major version update brings independently installable openqaoa plugins


Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+.PHONY: local-install
+	pip install ./src/openqaoa-core
+	pip install ./src/openqaoa-qiskit
+	pip install ./src/openqaoa-pyquil
+	pip install ./src/openqaoa-braket
+	pip install ./src/openqaoa-azure
+	pip install .
+
+.PHONY: dev-install
```
Q-lds, line 12: What is the difference between local-install and dev-install?

src/openqaoa-core/tests/test_workflows.py
shahidee44, line 1: https://github.com/entropicalabs/openqaoa/blob/b90589b3156982366be400d1f77170d5da840b0c/src/openqaoa-core/tests/test_workflows.py#L199

This line should be `qaoa_uncompiled.solve_brute_force()`

Makefile
vishal-ph: `local-install` installs the git cloned package as is, whereas `dev-install` installs the package as editable

shahidee44 (approved):

PR #264, opened 2023-07-14
fixed qubit overflow error by placing error catching mechanism before…

vishal-ph:
## Description

- If there are more number of qubits in the problem than are available on the device, OpenQAOA should raise an error before the problem is sent for routing to the routing function. This check was happening after the routing function was called. This PR fixes that and places the qubit number check before the routing function is called.

### DIFF
--- a/src/openqaoa-core/openqaoa/algorithms/baseworkflow.py
+++ b/src/openqaoa-core/openqaoa/algorithms/baseworkflow.py
@@ -376,7 +376,8 @@ class Workflow(ABC):
         if hasattr(self.device, "n_qubits"):
             if self.device.n_qubits < self.problem.n:
                 raise Exception(
-                    "There are lesser qubits on the device than the number of qubits required for the circuit."
+                    f"The number of qubits {self.problem.n} is more than the number of qubits available on the device."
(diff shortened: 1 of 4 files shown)


src/openqaoa-core/openqaoa/algorithms/baseworkflow.py
```diff
@@ -373,6 +373,12 @@ def compile(self, problem: QUBO):
         assert isinstance(problem, QUBO), "The problem must be converted into QUBO form"
         self.problem = problem
 
+        if hasattr(self.device, "n_qubits"):
+            if self.device.n_qubits < self.problem.n:
+                raise Exception(
+                    "There are lesser qubits on the device than the number of qubits required for the circuit."
```
Q-lds, line 379: I always find it useful seeing the offending numbers :D 

"The number of qubits {n_qubit} is lesser than the number of qubits available on the device: f{self.device.name} features f{self.device.n_qubits} qubits"

vishal-ph: done

Q-lds (approved):

PR #270, opened 2023-08-16
Dev spam twirling evaluate

KilianPoirier:
## Description

- I fixed a known issue that the SPAM twirling feature has not implemented `expectation_with_uncertainty()`, calling `evaluate_circuit()` would not call the wrapper's redefinition of `expectation()` with the previous version of this code.
- The function `evaluate_circuit()` now functions correctly, further implementation of the uncertainty calculation with SPAM twirling is still possible.

## Checklist

[/]: <> (- [ ] Any dependent changes have been merged and published in downstream modules)

## Type of change

### DIFF
--- a/src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_workflow.py
+++ b/src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_workflow.py
@@ -402,7 +402,7 @@ class QAOA(Workflow):
         # Check the type of the input parameters and save them as a
         # QAOAVariationalBaseParams object at the variable `params_obj`
 
-        # if the parameters are passed as a dictionary we copy and update the variational parameters of the QAOA object
+            # if the parameters are passed as a dictionary we copy and update the variational parameters of the QAOA object
         if isinstance(params, dict):
             params_obj = deepcopy(self.variate_params)
             # we check that the dictionary contains all the parameters of the QAOA object that are not empty
@@ -450,6 +450,20 @@ class QAOA(Workflow):
         # if the backend is the analytical simulator, we just return the expectation value of the cost Hamiltonian
         if isinstance(self.backend, QAOABackendAnalyticalSimulator):
             output_dict.update({"cost": self.backend.expectation(params_obj)[0]})
+        # if the workflow implements SPAM Twirling, uncertainty is not implemented
+        elif isinstance(self.backend, SPAMTwirlingWrapper):
+            cost = self.backend.expectation(params_obj)
+            measurement_results = (
+                self.backend.measurement_outcomes
+                if isinstance(self.backend.measurement_outcomes, dict)
+                else self.backend.measurement_outcomes.tolist()
+            )
+            output_dict.update(
+                {
+                    "cost": cost,
+                    "measurement_results": measurement_results,
+                }
+            )
 
         else:
             cost, uncertainty = self.backend.expectation_w_uncertainty(params_obj)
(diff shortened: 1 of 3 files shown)


src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_workflow.py
```diff
@@ -450,6 +450,21 @@ def evaluate_circuit(
...
+        elif isinstance(self.backend, SPAMTwirlingWrapper):
+            cost = self.backend.expectation(params_obj)
+            measurement_results = (
+                self.backend.measurement_outcomes
+                if isinstance(self.backend.measurement_outcomes, dict)
+                else self.backend.measurement_outcomes.tolist()
+            )
+            output_dict.update(
+                {
+                    "cost": cost,
```
vishal-ph, line 464: let's change the key `cost` to `expectation` in the `output_dict` for consistency with the docstring of the function

src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_workflow.py
```diff
@@ -450,6 +450,21 @@ def evaluate_circuit(
         # if the backend is the analytical simulator, we just return the expectation value of the cost Hamiltonian
         if isinstance(self.backend, QAOABackendAnalyticalSimulator):
             output_dict.update({"cost": self.backend.expectation(params_obj)[0]})
+        # if the workflow implements SPAM Twirling, 
+        # we just return the expectation value of the cost Hamiltonian and measurement outcomes
+        elif isinstance(self.backend, SPAMTwirlingWrapper):
+            cost = self.backend.expectation(params_obj)
+            measurement_results = (
+                self.backend.measurement_outcomes
```
vishal-ph, line 458: also,` measurement_results = ...` is repeated, so we can put it outside the if/else block

src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_workflow.py
KilianPoirier: Wouldn't it make more sense to change the docstring t `cost` instead?
I checked the implications of this change and the docstring is the only bit that mentions "uncertainty" as a field, or do you think the name `expectation` would be much clearer

src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_workflow.py
vishal-ph: @KilianPoirier, any updates on this?

src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_workflow.py
vishal-ph: Either is fine! You can make the choice, the main point is just consistency

src/openqaoa-core/openqaoa/algorithms/qaoa/qaoa_workflow.py
KilianPoirier: `measurement_result` has to be calculated after calling `expectation()` or `expectation_with_uncertainty()` which means that combining the two would result in either an additional `if`, `else` statement or an additional dictionary update. Here we can clearly distinguish that for different backends we calculate different quantities and the result dictionary is only updated once.

vishal-ph (approved):

PR #311, opened 2024-05-29
Add support to python 3.11

rodolfocarobene:
Fixes #307.

## Description

For the moment I've just upgraded all the setup.py and the workflows.
I will remove this PR from draft mode as soon as I'm finished with the changes. In particular, I want to check that the code is all compatible with python3.11 or if modifications are required.
I've yet to check the correctness of the dependencies in the opeqaoa-core: they may be outdated or require changes with the python version

## Checklist

[//]: <> (- [ ] My code follows the style guidelines of this project)

- [ ] I have performed a self-review of my code.
- [ ] I have made corresponding updates to the documentation.
- [ ] My changes generate no new warnings
- [ ] I have added/updated tests to make sure bugfix/feature works.
- [ ] New and existing unit tests pass locally with my changes.

[//]: <> (- [ ] Any dependent changes have been merged and published in downstream modules)

## Type of change

Please delete options that are not relevant.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

PR #312, opened 2024-05-30
migrated `openqaoa-qiskit` to qiskit 1.0

### LINKED ISSUE #304: Update `openqaoa-qiskit` to Qiskit 1.0
### Issue Description

OpenQAOA is currently not compatible with the latest major release of Qiskit 1.0 and above, for more details see [Qiskit release notes](https://docs.quantum.ibm.com/api/qiskit/release-notes/1.0).

It is important that we update the package requirements and migrate to this new release before the anterior version becomes obsolete and support from the IBM backends stop for older versions of the package.

This issue concerns mostly the Qiskit plugin (`openqaoa-qiskit`) but also has implications on the Azure plugin (`openqaoa-azure`) because of its dependencies.

### Changes to be made

- Update Qiskit dependencies in `openqaoa-qiskit` to be compatible with the latest major release. More details can be found in the [release notes](https://docs.quantum.ibm.com/api/qiskit/release-notes/1.0) and the [migration guide](https://docs.quantum.ibm.com/api/migration-guides/qiskit-1.0-installation) provided by IBM. Changes include:
    - Repackaging of the `qiskit` package. `qiskit-terra` has been removed and the core of the package is now located in `qiskit` itself. `qiskit-aer` now has to be imported on its own.
    - Update of the primitives features.
    - QASM conversion methods have been removed.
- Update Qiskit dependencies in `openqaoa-azure` if needed. Currently the version of Qiskit in the Azure quantum package is limited to `<1.0` but it may be updated in the near future.

prakharb10:
## Description

Closes #304

## Checklist

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

### DIFF
--- a/src/openqaoa-azure/requirements.txt
+++ b/src/openqaoa-azure/requirements.txt
@@ -1,4 +1,4 @@
 qdk
-qiskit-qir
+qiskit-qir~=0.5.0
 qiskit-ionq
 azure-quantum[qiskit]
--- a/src/openqaoa-qiskit/openqaoa_qiskit/backends/qaoa_qiskit_sim.py
+++ b/src/openqaoa-qiskit/openqaoa_qiskit/backends/qaoa_qiskit_sim.py
@@ -3,8 +3,8 @@ from typing import Union, List, Tuple, Optional
 
 # IBM Qiskit imports
 from qiskit import QuantumCircuit, QuantumRegister
-from qiskit.providers.aer import AerSimulator
-from qiskit.providers.aer.noise import NoiseModel
+from qiskit_aer import AerSimulator
+from qiskit_aer.noise import NoiseModel
 from qiskit.opflow.primitive_ops import PauliSumOp
 from qiskit.quantum_info import Statevector
 from qiskit.circuit import Parameter
--- a/src/openqaoa-qiskit/requirements.txt
+++ b/src/openqaoa-qiskit/requirements.txt
@@ -1,3 +1,3 @@
-qiskit>=0.36.1,<1.0
+qiskit~=1.0.0
 qiskit-ibm-provider
 qiskit-aer


KilianPoirier:
Hi @prakharb10 , thanks for filing a PR to solve this issue.

Can you change the target of the merge to `dev`, I'll review the changes and approve the workflow once this is changed.

prakharb10:
Hi @KilianPoirier 

I've switched the target branch to `dev`. The tests passed locally, but some didn't run due to a lack of credentials (I'm not sure about the reason, though). I'm happy to fix any failing tests in the workflow

KilianPoirier:
Seems like some of the changes you made were ignored when merging `dev` back into your branch.
It also seems like `qiskit.opflow` was not migrated see [migration guide](https://docs.quantum.ibm.com/api/migration-guides/qiskit-opflow-module).

prakharb10:
A sub-error originated from `qiskit-ibm-runtime` - https://github.com/Qiskit/qiskit-ibm-runtime/issues/1577. 
I added `pydantic` as a requirement to fix it but that is causing errors in other packages to show up. Trying to relax the version requirements

KilianPoirier:
I see, I have experienced warnings with the. versioning of `pydantic` before but it has never caused issues in installing any of the plugins. Hopefully, this is not becoming a major breaking point.

I noticed your comment on migrating to `ibm-runtime`. If this is something you want to look into, I would happy to support it. I know that this migration would be a big endeavour and this is why we have not pursued it at the moment. 

I haven't had a look at `qiskit-runtime` in quite a while so the compatibility with our codebase may have changed. We developed a[ light version of openqaoa using runtime primitives](https://github.com/entropicalabs/openqaoa-qiskit-lite-runtime) some time ago. You can check it out and see for yourself.

Let me know what you think of it!

prakharb10:
The test workflow is run on Python `3.8`. I've been running tests on Python `3.10` so far and could not reproduce the errors (though I did receive some other errors). Should the tests be run for all Python versions for which the package is built?

This could be why `pydantic` is not playing well across dependencies. I'm looking into it more.

---

I can look into migrating to `qiskit-ibm-runtime` in a separate PR to keep this clean.

KilianPoirier:
You are right, I overlooked this detail, the external dev test workflow is indeed only run on python 3.8. The rest of the workflow is actually run on all supported versions.

I was discussing dropping support for 3.8 with another unitary hacker, see #311 .

On your side you should still ensure that `openqaoa` can be installed and the tests pass at least on the olest and most recent versions.
If you come across cryptic error messages feel free to discuss them here.

prakharb10:
The errors popping up in my local run of the tests were related to `mitiq`, which I believe is because they no longer support `3.8`.
Switching to Python `3.9` as the minimum supported version sounds good. I can wait a day or two for #311 to merge and then rebase to see where we stand with the failing tests.

KilianPoirier:
So the tests are all passing locally?

prakharb10:
Sorry, I missed your message.

Not precisely, I have an M2 MacBook Air, and when I run `make dev-install-tests`, it fails with the message:

```
ERROR: Ignored the following versions that require a different python version: 1.14.0rc1 Requires-Python >=3.10; 3.3 Requires-Python >=3.10; 3.3rc0 Requires-Python >=3.10; 8.19.0 Requires-Python >=3.10; 8.20.0 Requires-Python >=3.10; 8.21.0 Requires-Python >=3.10; 8.22.0 Requires-Python >=3.10; 8.22.1 Requires-Python >=3.10; 8.22.2 Requires-Python >=3.10; 8.23.0 Requires-Python >=3.10; 8.24.0 Requires-Python >=3.10; 8.25.0 Requires-Python >=3.10
ERROR: Could not find a version that satisfies the requirement cplex>=22.1.0.0; extra == "tests" (from openqaoa-core[tests]) (from versions: none)
ERROR: No matching distribution found for cplex>=22.1.0.0; extra == "tests"
```

It appears that `cplex` does not have a supporting distribution. So far, I have been running it on a codespace to overcome this, but I can't change the Python version easily.

I've updated the workflow file to run on `3.9`. I'll see what tests fail now and work through them.

KilianPoirier:
It seems like there are some issues with test collection. Try to pull the changes I made on `dev` recently to prevent issues coming from mitiq.

KilianPoirier:
I have noticed the issue with `cplex` as well, it seems to not have a valid distribution on M2 chips. I'm looking into the problem but it doesn't seem to have a satisfying solution for now.

If you can run it on a codespace, it should not cause any issue because it seems to be working fine on Ubuntu.

src/openqaoa-core/requirements.txt
```diff
@@ -10,4 +10,4 @@ semantic_version>=2.10
 autoray>=0.3.1
 requests
 ipython>=8.2.0
-mitiq
+mitiq[qiskit]
```
KilianPoirier, line 13: Is the qiskit extension required? The point of having plugins is that you don't have to install all of the larger packages when you want to use openqaoa. Here we would introduce qiskit in the core package again.

KilianPoirier (commented):
Looks good to me! Only one question about the qiskit extension

src/openqaoa-core/requirements.txt
prakharb10: TL;DR: `ply` is a runtime requirement of `cirq-core[contrib]`/`mitiq[qiskit]`. We can add`ply` as a direct dependency or use `mitiq[qiskit]`.

---

The need for this originated from a test on notebook `examples/15_Zero_Noise_Extrapolation.ipynb`. This [run](https://github.com/entropicalabs/openqaoa/actions/runs/9411305493/job/25993668893) shows the full stack trace.
The dependency graph essentially is:
`openqaoa-core` -> `mitiq` -> `mitiq_qiskit` -> `cirq-core[contrib]`

I'm open to resolving this any other way you prefer.

KilianPoirier: This is a problem because ideally, we want to have all of our plugins separate, let's say if I want to install `openqaoa-core` on its own, I should be able to use all of its functionalities. 
The issue here is that if you add qiskit dependencies to the core package then it loses its purpose.

Can you try adding `ply` to the requirements locally to see if everything still works? Otherwise I will be in touch with the team that implemented ZNE to see if we can package it better to keep the plugin structure.

prakharb10: Well, the issue arose from the example notebook, not any of the package tests.

One more alternative could be to add `mitiq[qiskit]` as a requirement to `openqaoa-qiskit`. This seems logical and would add those required dependencies when using `openqaoa` with `qiskit`.

Let me know if that sounds reasonable, and I will make the change.

KilianPoirier: That sounds reasonable to me let's go with this solution!

KilianPoirier (approved):
Looks good to me!

I can't seem to be able to assign you to the issue. Can you try commenting the issue you solved and I'll assign you to the bounty then.

